PR #53, opened 2022-01-25
Make `WalletConnectURI` public

hewigovens:
We need to pass around `WalletConnectURI` before paring and typed uri is always safer than `string`, also it might better to have a `public func pair(uri: WalletConnectURI)` method in `WalletConnectClient`

### DIFF
--- a/Sources/WalletConnect/Types/WalletConnectURI.swift
+++ b/Sources/WalletConnect/Types/WalletConnectURI.swift
@@ -1,11 +1,11 @@
 import Foundation
 
-struct WalletConnectURI: Equatable {
+public struct WalletConnectURI: Equatable {
     
-    let topic: String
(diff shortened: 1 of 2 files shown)


hewigovens:
Add code sample for a better understanding this issue: 

```swift
public enum WCVersionedSession: Equatable {
     case v1(WCSession)
     case v2(WalletConnectURI)

     public static func from(string: String) -> WCVersionedSession? {
         if let session = WCSession.from(string: string) {
             return .v1(session)
         } else if let uri = WalletConnectURI(string: string) {
             return .v2(uri)
         }
         return nil
     }
 }
```

hewigovens:
@MisterVants  @llbartekll please check the sample code above

llbartekll (approved):

PR #62, opened 2022-02-01
#36 Upgrade method permissions checks

### LINKED ISSUE #36: Add state and permission checks for upgrade call
Making a session update call should trigger session state and permission checks on the sender AND the receiver sides.

MisterVants:
closes #36

Implements state, permission and controller checks on upgrade method for the sender and receiver.


Sources/WalletConnect/WalletConnectError.swift
```diff
@@ -2,19 +2,18 @@
 
 import Foundation
 
+// TODO: Migrate protocol errors to ReasonCode enum over time. Use WalletConnectError for client errors only.
```
llbartekll, line 5: 👍

llbartekll (approved):

PR #71, opened 2022-02-09
Remove client name

llbartekll:
because we have removed the controller property, proxy apps can use one instance of a client, as a result we no longer need a client name


dukevantreta (approved):

Doranayeli (commented):
malos

Doranayeli (commented):
malos

PR #77, opened 2022-02-14
WalletConnectError type cleanup

MisterVants:
Cleans the majority of unnecessary error cases in `WalletConnectError`, most of which were replaced by reason codes for p2p protocol errors.


Sources/WalletConnect/Engine/PairingEngine.swift
```diff
@@ -78,11 +75,8 @@ final class PairingEngine {
     
     func approve(_ pairingURI: WalletConnectURI) throws {
```
llbartekll, line 76: looks like this function can no longer throw

Sources/WalletConnect/Engine/SessionEngine.swift
```diff
@@ -196,16 +185,19 @@ final class SessionEngine {
...
+        guard session.isSettled else {
+            throw WalletConnectError.sessionNotSettled(topic)
+        }
+        guard session.selfIsController else {
+            throw WalletConnectError.unauthorizedNonControllerCall
         }
         for account in accounts {
             if !String.conformsToCAIP10(account) {
-                throw WalletConnectError.internal(.notApproved) // TODO: Use a suitable error cases
+                throw WalletConnectError.invalidCAIP10Account(account)
```
llbartekll, line 198: if we use `Set<Account>` as `update()` param instead of `Set<String>` we won't have to check it.

Sources/WalletConnect/WalletConnectError.swift
```diff
@@ -1,149 +1,50 @@
...
-
-// TODO: Migrate protocol errors to ReasonCode enum over time. Use WalletConnectError for client errors only.
 enum WalletConnectError: Error {
     
+    case pairingProposalFailed(Error)
+    case malformedPairingURI
+    case noPairingMatchingTopic(String)
     case noSessionMatchingTopic(String)
     case sessionNotSettled(String)
+    case invalidCAIP10Account(String)
```
llbartekll, line 8: this could be removed then if we use Account type as mentioned above

llbartekll (changes requested):
overall looks good, just a few comments

Sources/WalletConnect/Engine/SessionEngine.swift
MisterVants: Maybe it's even better if the `Account` object is stored in session sequences instead of strings.

Sources/WalletConnect/WalletConnectError.swift
MisterVants: Error case removed

PR #81, opened 2022-02-21
#8 extend method

### LINKED ISSUE #8: add extend method to increase expiry (controller-only)
client should have a public `extend()` method that allows to increase sequence expiry time.
only controller should be allowed to perform this request.
by default session should be extended by 7 days and pairing by 30 days

 
- [x]  Add extend method to pairing engine for controllers
- [x]  Add extend method to session engine for controllers
- [x]  Handle incoming extension internally in the pairing engine
- [x]  Handle incoming extension internally in the session engine
- [x]  add didExtend(pairing:) delegate method  - non controller
- [x]  add didExtend(session:) delegate method  - non controller
- [x]  Add error message for invalid or unauthorized extension
- [x]  Test happy path of valid extend method being called
- [x]  Test error message when non-controller calls extend
- [x]  Test error message when ttl is higher than default
- [x]  Test error message when ttl is lower than current expiry

llbartekll:
close #8 
does not contain tests for pairing extend - will be added if needed after protocol change decision


Sources/WalletConnect/WalletConnectClient.swift
```diff
@@ -163,6 +162,18 @@ public final class WalletConnectClient {
         try sessionEngine.upgrade(topic: topic, permissions: permissions)
     }
     
+    /// For controller to extend a sequence lifetime
+    /// - Parameters:
+    ///   - topic: Topic of the sequence, it can be a pairing or a session topic.
+    ///   - ttl: Time in seconds that a target sequence is expected to be extended for. Must be greater than current time to expiry and for session lower than 7 days and for pairing lower than 30 days.
+    public func extend(topic: String, ttl: Int) throws {
```
MisterVants, line 169: In the future we could add a constrained time enum as a parameter instead of the absolute time in seconds to make it simpler to call, like `extendBy: .month`

Tests/WalletConnectTests/PairingEngineTests.swift
```diff
@@ -148,17 +148,13 @@ final class PairingEngineTests: XCTestCase {
...
-//    func testNotifyOnSessionProposal() {
-//        let topic = "1234"
-//        let proposalExpectation = expectation(description: "on session proposal is called after pairing payload")
-////        engine.sequencesStore.create(topic: topic, sequenceState: sequencePendingState)
-//        try? engine.sequencesStore.setSequence(pendingPairing)
-//        let subscriptionPayload = WCRequestSubscriptionPayload(topic: topic, clientSynchJsonRpc: sessionProposal)
-//        engine.onSessionProposal = { (_) in
-//            proposalExpectation.fulfill()
+//
+//    func testNonControllerExtendFails() {
```
MisterVants, line 152: Are you going to reactivate or remove this test?

MisterVants (approved):

Tests/WalletConnectTests/PairingEngineTests.swift
llbartekll: I've removed it. We can test here when the "new pairing" is more defined

PR #99, opened 2022-02-28
refactor session proposal model

llbartekll:
- refactor session proposal model
- extract blockchains from permissions is session object
- remove ttl and default to session for 7 days

### DIFF
--- a/Sources/WalletConnect/Session.swift
+++ b/Sources/WalletConnect/Session.swift
@@ -10,6 +10,7 @@ public struct Session {
     public let permissions: Permissions
     public let accounts: Set<Account>
     public let expiryDate: Date
+    public let blockchains: Set<String>
 }
 
 extension Session {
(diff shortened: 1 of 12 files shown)


Sources/WalletConnect/Session.swift
```diff
@@ -10,31 +10,30 @@ public struct Session {
     public let permissions: Permissions
     public let accounts: Set<Account>
     public let expiryDate: Date
+    public let blockchains: Set<String>
```
MisterVants, line 13: What do you think of the idea of adding a CAIP-2 `Blockchain` type in the future, similar to `Account`? Does is make sense?

Sources/WalletConnect/Types/Session/SessionProposal.swift
```diff
@@ -6,27 +6,23 @@ struct SessionState: Codable, Equatable {
...
+    let publicKey: String
+    let controller: Bool
+    let metadata: AppMetadata
+}
+//?
+struct BlockchainProposed: Codable, Equatable {
+    //TODO - auth type not specified yet
+    let auth: String? = nil
+    // TODO - change for caip2 objects
+    let chains: Set<String>
```
MisterVants, line 27: Ignore the last comment, you got the same idea

MisterVants (approved):

PR #100, opened 2022-03-01
Propose session

llbartekll:
add propose method

### DIFF
--- a/Sources/WalletConnect/Engine/SessionEngine.swift
+++ b/Sources/WalletConnect/Engine/SessionEngine.swift
@@ -59,40 +59,28 @@ final class SessionEngine {
         }
     }
         
-    func proposeSession(settledPairing: Pairing, permissions: SessionPermissions, relay: RelayProtocolOptions) {
-        guard let pendingSessionTopic = topicInitializer() else {
-            logger.debug("Could not generate topic")
-            return
-        }
-        logger.debug("Propose Session on topic: \(pendingSessionTopic)")
-        
+    func proposeSession(pairing: Pairing, permissions: SessionPermissions, relay: RelayProtocolOptions) {
+        logger.debug("Propose Session on topic: \(pairing.topic)")
         let publicKey = try! kms.createX25519KeyPair()
-        
+        let proposer = Proposer(
+            publicKey: publicKey.hexRepresentation,
+            metadata: metadata)
         let proposal = SessionProposal(
-            topic: pendingSessionTopic,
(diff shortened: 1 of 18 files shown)


Sources/WalletConnect/Engine/SessionEngine.swift
```diff
@@ -282,46 +246,37 @@ final class SessionEngine {
...
-        guard !session.selfIsController else {
-            // TODO: Replace generic reason with a valid code.
-            relayer.respondError(for: payload, reason: .generic(message: "wcSessionApproval received by a controller"))
+    private func wcSessionPropose(_ payload: WCRequestSubscriptionPayload, proposal: SessionType.ProposeParams) {
+        proposerToRequestPayload[proposal.proposer.publicKey] = payload
+        onSessionProposal?(proposal)
+    }
+        
+    func respondSessionPropose(proposal: SessionType.ProposeParams) {
+        guard let payload = proposerToRequestPayload[proposal.proposer.publicKey] else {
```
MisterVants, line 255: Don't you need to clear the cached proposal after responding?

Sources/WalletConnect/Engine/SessionEngine.swift
```diff
@@ -282,46 +246,37 @@ final class SessionEngine {
...
-        logger.debug("Responder Client approved session on topic: \(topic)")
-        relayer.respondSuccess(for: payload)
-        onSessionApproved?(approvedSession)
+        sequencesStore.setSequence(pendingSession)
+
+        try! kms.setAgreementSecret(agreementKey, topic: sessionTopic)
+        wcSubscriber.setSubscription(topic: sessionTopic)
+        let proposeResponse = SessionType.ProposeResponse(relay: proposal.relay, responder: AgreementPeer(publicKey: selfPublicKey.hexRepresentation))
+        let response = JSONRPCResponse<AnyCodable>(id: payload.wcRequest.id, result: AnyCodable(proposeResponse))
+        relayer.respond(topic: sessionTopic, response: .response(response)) { _ in }
```
MisterVants, line 279: I think you're responding the proposal over the wrong topic. The proposal must be responded on topic A, while the settlement request that is sent on topic B.

Sources/WalletConnect/Engine/SessionEngine.swift
MisterVants: I don't like it either, let's remove it later

MisterVants (approved):

PR #105, opened 2022-03-02
Serializer refactor

### LINKED ISSUE #101: Allow to encrypt messages with any symKey

llbartekll:
Allows to encrypt messages with agreement key and symmetric key. close #101


Sources/WalletConnect/KeyManagementServiceProtocol.swift
```diff
@@ -2,20 +2,5 @@
 import Foundation
```
MisterVants, line 2: This file is now empty, can we just delete it?

llbartekll: done

Issue #18, opened 2022-01-11
Test expiration times on all state sequences


MisterVants:
Expiration times should be tested for pairings and sessions, not only in settled states, but also in intermediate states (non-responded, pre-settled, etc).

PR #130, opened 2022-03-22
#18 test pairing deletion

MisterVants:


Tests/WalletConnectTests/PairingEngineTests.swift
```diff
@@ -164,12 +164,39 @@ final class PairingEngineTests: XCTestCase {
...
+        
+        guard let request = relayMock.requests.first?.request,
+              let proposal = relayMock.requests.first?.request.sessionProposal else {
+                  XCTFail("Proposer must publish session proposal request"); return
+              }
+        
+        let response = WCResponse.stubError(forRequest: request, topic: topicA)
+        relayMock.onPairingResponse?(response)
+        
+        XCTAssert(subscriberMock.didUnsubscribe(to: pairing.topic), "Proposer must not unsubscribe if pairing is inactive.")
```
llbartekll, line 191: must not?

Sources/WalletConnect/Engine/PairingEngine.swift
```diff
@@ -198,8 +199,18 @@ final class PairingEngine {
...
+            return
+        }
         switch result {
         case .response(let response):
+            
+            // Activate the pairing
+            if !pairing.isActive {
+                pairing.activate()
+                sequencesStore.setSequence(pairing)
+            }
```
llbartekll, line 212: if it is active then we need to extend

llbartekll (approved):

PR #150, opened 2022-04-07
#18 Pairing expiry tests

MisterVants:
- Add tests for pairing expiry dates and updates.
- Replaces legacy factory method from when a pairing had multiple sub-states.
- Binds extended expiry with activation.

### DIFF
--- a/Tests/WalletConnectTests/PairingSequenceTests.swift
+++ b/Tests/WalletConnectTests/PairingSequenceTests.swift
@@ -1,0 +1,32 @@
+import XCTest
+@testable import WalletConnect
+
+final class PairingSequenceTests: XCTestCase {
+    
+    var referenceDate: Date!
+    
+    override func setUp() {
+        referenceDate = Date()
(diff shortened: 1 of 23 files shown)


Tests/WalletConnectTests/PairingSequenceTests.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+    
+    func testInitInactiveFromURI() {
+        let pairing = PairingSequence(uri: WalletConnectURI.stub())
+        let inactiveExpiry = referenceDate.advanced(by: PairingSequence.timeToLiveInactive)
+        XCTAssertFalse(pairing.isActive)
+        XCTAssertEqual(pairing.expiryDate, inactiveExpiry)
+    }
+    
+    func testExtend() {
```
llbartekll, line 37: testUpdateExpiry() instead? we could stop using extend keyword

MisterVants: Done

PR #155, opened 2022-04-11
Add Commons and Toolbox packages

MisterVants:
Introduces 2 base dependency packages: Commons and Toolbox, and extract some common code to them. They are base dependencies for the SDK and the tests target. We could follow some guidelines for these packages to help maintaining the health of the project on the long term.

### Commons:
- Base dependency for all other packages of the SDK. 
- Must not have any dependency itself. 
- Must not contain code that makes reference to any domain-specific logic (JSON-RPC, storage, engine, etc)
- Should stay as lean as possible and contain only essential data types and extensions (like `AnyCodable`).
- Everything (ideally) should be tested.

### Toolbox:
- Base dependency for test utilities.
- Should not have any dependencies (unless it really saves code duplication, if we confirm it later).
- We should avoid domain-specific test doubles (mocks), these can remain in their targets.
- Should only have code that is used in multiple test targets.

### Notes:
1. Noticing that in WalletConnectUtils there's domain-specific code, it will be split into core components in the future and can be removed when not needed anymore.
2. Toolbox is a cool name. Commons is still a simple placeholder name, we should try to find a better name to brand the package in the future.
3. `Either` type was re-introduced, it will help build a full JSON-RPC compliant API for requests.


llbartekll:
@MisterVants I think we can create readme for each package and add your guidelines from PR there.

MisterVants:
TODO: Rename `TestingUtils` to `Toolbox` and try to reduce dependencies to a single package.

MisterVants:
> TODO: Test usage of sub-package types with a single `import` statement
> 
> (needs to be done before following up with the new structure)

Solution 1: Use typealiases to expose sub-package public types. A list can be maintained in a single file.
Solution 2: Use umbrella frameworks. Might not be the best approach: https://stackoverflow.com/questions/7365578/why-are-umbrella-frameworks-discouraged

MisterVants:
Commons package added in PR #256, together with JSONRPC package. TestingUtils to be renamed to Toolbox later.

PR #160, opened 2022-04-14
#159 Fix example apps build

MisterVants:


llbartekll (changes requested):
I have reenabled example apps automatic builds and they fail.

PR #175, opened 2022-04-27
Update README documentation

MisterVants:
Updates the README documentation to reflect the latest SDK changes.


llbartekll (commented):
I feel like we discussed that we no longer want to keep sample code in readme but in common docs instead.

PR #187, opened 2022-05-02
Update issue templates

chadyj:
Create Github Issue templates similar to the monorepo.

https://github.com/WalletConnect/walletconnect-monorepo/issues/new/choose

Screenshot 2022-05-02 at 7 50 26 PM


MisterVants:
We could use an extra template for simple To-Do's, to describe better things that aren't bugs or requests.

PR #194, opened 2022-05-05
fix wallet to not operate on mocked data

llbartekll:
quick fix to not operate on mocked data and allow manual testing


MisterVants (approved):
Still needs to fix mock, going to do on next PR

PR #215, opened 2022-05-17
#211 Namespace validation

MisterVants:
Adds validation functions to new namespaces, applies planned test cases.

Integrates validation on proposal-approval flow.


Sources/WalletConnectAuth/Engine/Common/PairingEngine.swift
```diff
@@ -209,6 +212,7 @@ final class PairingEngine {
                 try? pairing.updateExpiry()
             }
             
+            onApprovalResponse?(proposal) // FIXME: This proposal needs to be stored on creation and retrieved from cache here. Don't use proposal from response.
```
llbartekll, line 215: `This proposal needs to be stored on creation and retrieved from cache here. Don't use proposal from response.`
actually that proposal is from cache(jsonRpcHistory record) not from response, response do not contain proposal.

Sources/WalletConnectAuth/AuthClient.swift
```diff
@@ -330,5 +330,8 @@ public final class AuthClient {
         pairingEngine.onProposeResponse = { [unowned self] sessionTopic in
             sessionEngine.setSubscription(topic: sessionTopic)
         }
+        pairingEngine.onApprovalResponse = { [unowned self] proposal in
+            sessionEngine.settlingProposal = proposal
+        }
```
llbartekll, line 335: do you need that extra callback `onApprovalResponse` ? `onProposeResponse` is executed in the same function.

would it work to have two arguments(if needed) in on callback?

Sources/WalletConnectAuth/Engine/Common/SessionEngine.swift
```diff
@@ -12,6 +12,8 @@ final class SessionEngine {
     var onSessionDelete: ((String, SessionType.Reason)->())?
     var onEventReceived: ((String, Session.Event, Blockchain?)->())?
     
+    var settlingProposal: SessionProposal?
```
llbartekll, line 15: could it be overriden?
could some runtime key value storage be safer?

Sources/WalletConnectAuth/Engine/Common/SessionEngine.swift
```diff
@@ -139,34 +165,23 @@ final class SessionEngine {
...
-        logger.debug("Sending session settle request")
-        Task { try? await networkingInteractor.subscribe(topic: topic) }
-        sessionStore.setSession(session)
-        networkingInteractor.request(.wcSessionSettle(settleParams), onTopic: topic)
-    }
-
+    
     private func onSessionSettle(payload: WCRequestSubscriptionPayload, settleParams: SessionType.SettleParams) {
         logger.debug("Did receive session settle request")
+        guard let proposedNamespaces = settlingProposal?.requiredNamespaces else {
```
llbartekll, line 171: 1. you set `settlingProposal` `onApprovalResponse` so it happens on `responder's` engine as responder calls "approve"
2. you retrieve that `settlingProposal` `onSessionSettle` so it happens on `proposer's` engine as responder calls  "settle"

🤔  am I correct? will it work?

llbartekll: splitting engines/methods for controller/non-controller later will really help.

llbartekll (changes requested):
nice namespaces validation and tests
take a look at my comments, we can discuss if you don't agree on something

Sources/WalletConnectAuth/Engine/Common/PairingEngine.swift
```diff
@@ -116,33 +117,35 @@ final class PairingEngine {
...
         }
+        let proposedNamespaces = proposal.requiredNamespaces
         proposalPayloadsStore.delete(forKey: proposerPubKey)
-
-        let selfPublicKey = try! kms.createX25519KeyPair()
-        var agreementKey: AgreementKeys!
         
         do {
-            agreementKey = try kms.performKeyAgreement(selfPublicKey: selfPublicKey, peerPublicKey: proposal.proposer.publicKey)
+            try Namespace.validate(proposedNamespaces)
```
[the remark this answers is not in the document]
MisterVants: Actually no, but it should. I'm changing it

Sources/WalletConnectAuth/Engine/Common/PairingEngine.swift
```diff
@@ -116,33 +117,35 @@ final class PairingEngine {
...
+            try Namespace.validate(sessionNamespaces)
+            try Namespace.validateApproved(sessionNamespaces, against: proposedNamespaces)
```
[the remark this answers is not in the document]
MisterVants: I agree with the validation error catch, it has to be thrown. I just used it as a temporary solution to make the validation start working. An issue can be created for that to improve error reading.

The catch respond error already existed before, but only when the key agreement failed. Should this be changed also?

Sources/WalletConnectAuth/AuthClient.swift
MisterVants: Yes, changed the callback to 2 args

llbartekll (approved):

PR #231, opened 2022-05-24
[Wallet] Edit session namespaces

flypaper0:
**What changed**
- replaced iconURL's because old doesn't load
- SessionDetails screen rewrited to SwiftUI
- Ability to delete namespace properties 

**Video** _(Sorry for quality, 10MB limit)_

https://user-images.githubusercontent.com/5384197/170071174-1a3e6546-634d-4d74-a130-89aa262d1dd2.mov


MisterVants:
We could use some protocol abstraction to add a full preview to this view in the future.

Example/ExampleApp/SessionDetails/SessionDetailViewModel.swift
```diff
@@ -0,0 +1,140 @@
+import Foundation
+import Combine
+import WalletConnectAuth
+
+@MainActor
```
MisterVants, line 5: Why does it need to be a `@MainActor`? Does it have something to do with setting a `@Published` var during an async call?

MisterVants (approved):

llbartekll (approved):

Example/ExampleApp/SessionDetails/SessionDetailViewModel.swift
flypaper0: Yes, ensure what UI update in main thread

flypaper0:
> We could use some protocol abstraction to add a full preview to this view in the future.

Anyway we need to mock SessionDetailViewModel somehow to display screen. Or what you mean?

PR #234, opened 2022-05-25
Hot fix dapp personal sign

llbartekll:
use session account for personal_sign request in order to allow testing with js dapp


Example/DApp/AccountRequest/AccountRequestViewController.swift
```diff
@@ -79,12 +79,12 @@ class AccountRequestViewController: UIViewController, UITableViewDelegate, UITab
     }
     
     private func getRequest(for method: String) -> AnyCodable {
-        let account = "0x9b2055d370f73ec7d8a03e965129118dc8f5bf83"
+        let account = session.namespaces.first!.value.accounts.first!.absoluteString
```
flypaper0, line 82: namespaces and accounts couldn't be empty?

flypaper0 (approved):

PR #242, opened 2022-05-27
UI tests

flypaper0:
**What changed**
- UITesting environment 
- TU001, TU002, TU004 cases covered
- Alert on Ping response

**Video**
https://drive.google.com/file/d/11uRbgL1gZprK_LVotNO2oKy2Ml6dX_kI/view?usp=sharing


llbartekll (approved):
@flypaper0 just FYI: tests fail to setup on my mac.

llbartekll:
more specifically apps deletion process

Mike117687:
merge-these-changes

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

flypaper0:
> more specifically apps deletion process

This is one of limitation of my approach. For now it only works with English OS language

chadyj:
I think it is a good idea to integrate this into CI because we can test on multiple devices and OS versions. This will increase our testing surface and make it more likely we will spot edge cases.


During discussions we said this can happen only when merging to main, but out of curiosity, why not run e2e tests on every PR?

chadyj:
Some suggestions from the cloud team:

Concurrency e.g. https://github.com/WalletConnect/registry-api-cf-workers/blob/master/.github/workflows/pr.yaml#L15
Personal preference: split up this CI/PR step into two distinct steps (gives you a clean status badge)


I think this can improve the build 
```
- uses: actions/cache@v2
  with:
    path: Pods
    key: ${{ runner.os }}-pods-${{ hashFiles('**/Podfile.lock') }}
    restore-keys: |
            ${{ runner.os }}-pods-
 ```
Here is a good post about it https://benoitpasquier.com/tips-tweaks-github-action-ios/

Mike117687:
merge-these-changes

PR #262, opened 2022-06-07
#chat/registry

llbartekll:
- Add registry manager
- update client api with specs
- add Concurrency

### DIFF
--- a/Sources/Chat/Registry.swift
+++ b/Sources/Chat/Registry.swift
@@ -9,11 +9,12 @@ protocol Registry {
 
 actor KeyValueRegistry: Registry  {
     
-    var registryStore: [Account: String] = [:]
+    private var registryStore: [Account: String] = [:]
     
     func register(account address: Account, pubKey: String) async throws {
         registryStore[address] = pubKey
     }
+    
     func resolve(account: Account) async throws -> String {
         guard let record = registryStore[account] else { throw ChatError.recordNotFound}
         return record
--- a/Sources/Chat/Registry.swift
+++ b/Sources/Chat/Registry.swift
@@ -3,8 +3,8 @@ import Foundation
 import WalletConnectUtils
 
 protocol Registry {
-    func register(account: Account, pubKey: String) async throws 
-    func resolve(account: Account) async throws -> String?
+    func register(account: Account, pubKey: String) async throws
+    func resolve(account: Account) async throws -> String
 }
 
 actor KeyValueRegistry: Registry  {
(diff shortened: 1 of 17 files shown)


Sources/Chat/RegistryManager.swift
```diff
@@ -0,0 +1,35 @@
+
+import Foundation
+import WalletConnectUtils
+import WalletConnectKMS
+
+actor RegistryManager {
```
flypaper0, line 6: What the difference between Manager and Engine naming?

Tests/ChatTests/RegistryManagerTests.swift
```diff
@@ -0,0 +1,45 @@
...
+        registryManager = RegistryManager(
+            registry: registry,
+            networkingInteractor: networkingInteractor,
+            kms: kms,
+            logger: ConsoleLoggerMock(),
+            topicToInvitationPubKeyStore: topicToInvitationPubKeyStore)
+    }
+    
+    override func tearDown() {
+        registry = nil
```
flypaper0, line 29: May be not necessary to set nil at `tearDown` ? We setting new value in `setUp`

Sources/Chat/Registry.swift
```diff
@@ -3,19 +3,20 @@ import Foundation
...
 
 protocol Registry {
-    func register(account: Account, pubKey: String)
-    func resolve(account: Account) -> String?
+    func register(account: Account, pubKey: String) async throws
+    func resolve(account: Account) async throws -> String
 }
 
-class KeyValueRegistry: Registry  {
+actor KeyValueRegistry: Registry  {
```
flypaper0, line 10: What benefits actor gives us here? 

We could make it class and make registryStore private. So it couldn't be mutated from outside too

flypaper0 (approved):
async usage looks nice and convenient to use

Sources/Chat/RegistryManager.swift
llbartekll: I am always scratching my head when giving them a names. in this context it just sounds better to me. Any suggestions?

Sources/Chat/Registry.swift
llbartekll: true it is better `registryStore` is private
but I think making it an actor gives thread safety

> So it couldn't be mutated from outside too

It could be mutated by calling register() from different threads
anyway this class is mimicking an http client right now

MisterVants:
This PR is pointing to `main` after the `develop` merge, need to fix it

PR #270, opened 2022-06-13
ApproveEngine async calls

flypaper0:
**What changed?**

**Sample app** 
Approve / Reject calls wrapped in Tasks. Errors handled by logger

**SDK**
- Async `Approve` and `Reject` methods
- `NetworkInteractor`: Async RespondError and RespondSuccess methods
- `Dispatcher`: Async Send method
- Async callbacks for `wcRequestPublisher` subscribers
- Respond Namespace validation errors

### DIFF
--- a/Example/ExampleApp/SessionDetails/SessionDetailViewController.swift
+++ b/Example/ExampleApp/SessionDetails/SessionDetailViewController.swift
@@ -36,18 +36,18 @@ final class SessionDetailViewController: UIHostingController<SessionDetailView> 
     }
     
     private func respondOnSign(request: Request, response: JSONRPCResponse<AnyCodable>) {
-        print("[CONTROLLER] Respond on Sign")
+        print("[WALLET] Respond on Sign")
         Task {
             do {
                 try await Sign.instance.respond(topic: request.topic, response: .response(response))
             } catch {
-                print("[NON-CONTROLLER] Respond Error: \(error.localizedDescription)")
+                print("[DAPP] Respond Error: \(error.localizedDescription)")
             }
         }
     }
     
     private func respondOnReject(request: Request) {
-        print("[CONTROLLER] Respond on Reject")
+        print("[WALLET] Respond on Reject")
         Task {
             do {
                 try await Sign.instance.respond(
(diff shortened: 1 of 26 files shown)


llbartekll (approved):

Example/ExampleApp/SessionDetails/SessionDetailViewController.swift
```diff
@@ -25,21 +25,43 @@ final class SessionDetailViewController: UIHostingController<SessionDetailView>
...
-                )
-            )
+            respondOnReject(request: request)
             reload()
         }
         present(viewController, animated: true)
     }
+    
+    private func respondOnSign(request: Request, response: JSONRPCResponse<AnyCodable>) {
+        print("[CONTROLLER] Respond on Sign")
```
MisterVants, line 39: We could already replace "[controller]" in logs with "[wallet]" wherever we find it. The controller terminology is a lot more implicit and internal to the SDK, now. I think just using "wallet" and "dapp" will make easier for external devs to understand the code.

Example/ExampleApp/SessionDetails/SessionDetailViewController.swift
```diff
@@ -25,21 +25,43 @@ final class SessionDetailViewController: UIHostingController<SessionDetailView>
...
+        Task {
+            do {
+                try await Sign.instance.respond(topic: request.topic, response: .response(response))
+            } catch {
+                print("[NON-CONTROLLER] Respond Error: \(error.localizedDescription)")
```
MisterVants, line 44: Same point for "non-controller" and "Dapp"

Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
```diff
@@ -69,66 +69,68 @@ final class ApproveEngine {
...
+            expiry: Int64(expiry))
+
         let session = WCSession(
             topic: topic,
             selfParticipant: selfParticipant,
             peerParticipant: proposal.proposer,
             settleParams: settleParams,
-            acknowledged: false
-        )
+            acknowledged: false)
```
MisterVants, line 126: I always ask myself whether to break a line or not when closing a parentheses 😅

MisterVants (approved):

Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
flypaper0: I almost always wrap to the next line. Don't remember why I decided to do this 😅

PR #274, opened 2022-06-16
[Wallet] State synchronisation

flypaper0:
**What changed**

- callbacks implemented in for `handleUpdateResponse` and `handleUpdateExpiryResponse`
- timestamp added to WCSession
- WCSession migration example (for timestamp)
- WCSessionStorage - isNeedToReplace logic
- ExpirableSequence renamed to SequenceObject
- SessionStorageTests

### DIFF
--- a/Sources/WalletConnectSign/Types/Session/WCSession.swift
+++ b/Sources/WalletConnectSign/Types/Session/WCSession.swift
@@ -2,8 +2,7 @@ import Foundation
 import WalletConnectKMS
 import WalletConnectUtils
 
-struct WCSession: ExpirableSequence {
-
+struct WCSession: SequenceObject, Equatable {
     enum Error: Swift.Error {
         case controllerNotSet
         case unsatisfiedUpdateNamespaceRequirement
@@ -13,9 +12,11 @@ struct WCSession: ExpirableSequence {
     let relay: RelayProtocolOptions
     let selfParticipant: Participant
     let peerParticipant: Participant
-    private (set) var expiryDate: Date
-    var acknowledged: Bool
     let controller: AgreementPeer
+
+    private(set) var acknowledged: Bool
+    private(set) var expiryDate: Date
+    private(set) var timestamp: Date
     private(set) var namespaces: [String: SessionNamespace]
     private(set) var requiredNamespaces: [String: SessionNamespace]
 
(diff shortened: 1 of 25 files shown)


Sources/WalletConnectSign/Types/Session/WCSession.swift
```diff
@@ -107,30 +113,33 @@ struct WCSession: ExpirableSequence {
...
-    mutating func updateExpiry(to expiry: Int64) throws {
+    mutating func updateExpiry(to expiry: Int64, timestamp: Date) throws {
         let newExpiryDate = Date(timeIntervalSince1970: TimeInterval(expiry))
         let maxExpiryDate = Date(timeIntervalSinceNow: TimeInterval(WCSession.defaultTimeToLive))
         guard newExpiryDate > expiryDate && newExpiryDate <= maxExpiryDate else {
             throw WalletConnectError.invalidUpdateExpiryValue
         }
-        expiryDate = newExpiryDate
+        self.expiryDate = newExpiryDate
+        self.timestamp = timestamp
```
llbartekll, line 142: so you decided to have one timestamp for `extend` and `update`....

Sources/WalletConnectUtils/JsonRecID.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+public struct JsonRpcID {
+
+    public static func generate() -> Int64 {
+        let timestamp = Int64(Date().timeIntervalSince1970 * 1000) * 1000
```
llbartekll, line 6: we probably should change this for real milliseconds as this are seconds multiplied by 1000 now.

llbartekll: it is likely to happen that there will be two requests send in one seconds and there is 50% chance that random suffix will be greater in first request. if it happens, second request will not update the session even though it should.

Sources/WalletConnectSign/Types/Session/WCSession.swift
```diff
@@ -107,30 +113,33 @@ struct WCSession: ExpirableSequence {
...
     }
 
     /// updates session expiry to given timestamp
     /// - Parameter expiry: timestamp in the future in seconds
-    mutating func updateExpiry(to expiry: Int64) throws {
+    mutating func updateExpiry(to expiry: Int64, timestamp: Date) throws {
```
llbartekll, line 135: should be changed already but .... `updateExpiry` ought to be replaced with `extend` where possible.
you can do it now or we can refactor it later

Sources/WalletConnectSign/Types/Session/WCSession.swift
flypaper0: Removed this logic for extendSession

llbartekll (approved):

Sources/WalletConnectSign/Types/Session/WCSession.swift
flypaper0: This will be a big diff. Created for that https://github.com/WalletConnect/WalletConnectSwiftV2/issues/282

PR #301, opened 2022-06-28
[Feature] Protocol renaming #272

flypaper0:
**What changed?**
- waku methods renamed to iridium
- URI format changed
- validate only last part of subscription method


MisterVants:
Just need to fix the CI check, the rest is good.

llbartekll (approved):

PR #363, opened 2022-07-25
[Refactor] SDK's Factories 

flypaper0:
**What changed**

ChatClientFactory adn SignClientFactory implemented. Client's init methods clean now

### DIFF
--- a/Example/Showcase/Classes/DomainLayer/Chat/ChatFactory.swift
+++ b/Example/Showcase/Classes/DomainLayer/Chat/ChatFactory.swift
@@ -2,6 +2,7 @@ import Foundation
 import Chat
 import WalletConnectKMS
 import WalletConnectRelay
+import WalletConnectUtils
 
 class ChatFactory {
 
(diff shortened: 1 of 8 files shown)


Sources/WalletConnectSign/Sign/SignClientFactory.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+import WalletConnectRelay
+import WalletConnectUtils
+import WalletConnectKMS
+
+public struct SignClientFactory {
+
+    /// Initializes and returns newly created WalletConnect Client Instance. Establishes a network connection with the relay
```
llbartekll, line 8: we can remove this part from comment:

> Establishes a network connection with the relay

llbartekll (approved):

Example/Showcase/Classes/DomainLayer/Chat/ChatFactory.swift
```diff
@@ -12,10 +13,11 @@ class ChatFactory {
         let client = HTTPClient(host: "keys.walletconnect.com")
         let registry = KeyserverRegistryProvider(client: client)
         let relayClient = RelayClient(relayHost: relayHost, projectId: projectId, keychainStorage: keychain, socketFactory: SocketFactory())
-        return ChatClient(
+        return ChatClientFactory.create(
```
llbartekll, line 16: I do not like that the integrator need to use factory
in Sign we were able to avoid this as we have a `Sign.instance` 
but we will probably add singleton wrapper for Chat as well

llbartekll (approved):

Example/Showcase/Classes/DomainLayer/Chat/ChatFactory.swift
flypaper0: yes, ChatClientFactory.create will be called in Chat singleton only

PR #369, opened 2022-07-26
[CI] Upgrade XCode on GH Runner

flypaper0:
**What Changed**
- macos version 12
- xcode version 13.4.1
- Chat integration tests uncommented

### DIFF
--- a/Example/IntegrationTests/Chat/ChatTests.swift
+++ b/Example/IntegrationTests/Chat/ChatTests.swift
@@ -47,7 +47,8 @@ final class ChatTests: XCTestCase {
     }
 
     func testInvite() async {
-        await waitClientsConnected()
+        waitClientsConnected()
+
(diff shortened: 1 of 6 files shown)


llbartekll (approved):

Example/IntegrationTests/Chat/ChatTests.swift
```diff
@@ -18,23 +18,23 @@ final class ChatTests: XCTestCase {
         inviter = makeClient(prefix: "🍄 Inviter")
     }
 
-    private func waitClientsConnected() async {
```
llbartekll, line 21: maybe instead of this method `waitClientsConnected` we could wait for clients connected in `setUp` function
look for sign integration tests for reference
but not super important

llbartekll (approved):

PR #381, opened 2022-07-28
#379 auth core types

llbartekll:
Add core types to Auth sdk

### DIFF
--- a/Sources/Auth/Types/AuthPayloadParams.swift
+++ b/Sources/Auth/Types/AuthPayloadParams.swift
@@ -1,0 +1,16 @@
+import Foundation
+
+struct AuthPayloadParams {
+    let type: String
+    let chainId: String
+    let domain: String
+    let aud: String
+    let version: String
+    let nonce: String
+    let iat: String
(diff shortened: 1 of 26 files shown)


Sources/Auth/Types/AuthPayloadParams.swift
```diff
@@ -0,0 +1,16 @@
...
+struct AuthPayloadParams {
+    let type: String
+    let chainId: String
+    let domain: String
+    let aud: String
+    let version: String
+    let nonce: String
+    let iat: String
+    let nbf: String?
+    let exp: String?
```
MisterVants, line 12: I still defend the approach of using readable names and declaring the abbreviated coding keys 😬

MisterVants (approved):

Sources/Auth/Types/AuthPayloadParams.swift
llbartekll: That’s according to cacao specs :-)

flypaper0 (approved):

PR #383, opened 2022-07-28
[Auth] Pairing Services

flypaper0:
**What changed?**
- CI env changed to lates stable
- WalletConnectPairing package
- AppPairService and Wallet PairService implemented
- Public interface for request and pair


Sources/Auth/Services/Wallet/AuthService.swift
```diff
@@ -1,6 +1,30 @@
...
+    }
+
+    func pair(uri: String) async throws {
+        guard let pairingURI = WalletConnectURI(string: uri) else {
+            throw Errors.malformedPairingURI
+        }
+        try await walletPairService.pair(pairingURI)
+    }
 
     func respond(respondParams: RespondParams) async throws {
```
llbartekll, line 29: My intention was that AuthService is for wallet to call `respond(respondParams: RespondParams)` method only.
`pair` and `request` are called from client directly
`request` calls `appPairService.create()`

take a look at client api specs:
https://github.com/WalletConnect/walletconnect-specs/pull/88/files#diff-6409c08fd6475bc986e81c120780838a754491571d419553faee490bb9922801

flypaper0: Moved to AuthClient

Sources/Auth/AuthClient.swift
```diff
@@ -1,5 +1,28 @@
...
+    private let appPairService: AppPairService
+
+    private let walletPairService: WalletPairService
+
+    init(appPairService: AppPairService, walletPairService: WalletPairService) {
+        self.appPairService = appPairService
+        self.walletPairService = walletPairService
+    }
+
+    func connect() async throws -> String {
```
llbartekll, line 17: if you look at specs, we do not have a `connect()` method. We have `request()` instead that is suppose to do as follows:
```
appPairService.create()
authRequestService().request
return uri
```

flypaper0: Like this? 
```
    func request(params: RequestParams) async throws -> String {
        let uri = try await appPairService.create()
        try await appRequestService.request(params: params, topic: uri.topic)
        return uri.absoluteString
    }
```

llbartekll (approved):

PR #384, opened 2022-07-28
[Network] Refactor: #355 Relay client RPC

### LINKED ISSUE #355: Refactor relayer's RPC methods
Refactor and scale RPC methods used in RelayClient and integrate the JSONRPC types:
- Subscribe
- Unsubscribe
- Publish
- Subscription

Mains goals should be to standardize and scale RPC methods to the products APIs and allow for easier maintenance of RPC layers.

MisterVants:
closes #355 

### Overview:
- Integration of `JSONRPC` types into the Relay client target.
- Introduces relay RPC methods declared conforming to `RelayRPC` protocol.
- Small improvements to `Either` and `RPCResponse`.
- ID generation logic for RPC methods decoupled in isolation.
- Introduces improved and expanded `RPCHistory` to track all RPC requests sent and received from the client.

### Notes:
- `RPCMethod` protocol could be extracted to utils to be used in other targets for scalability.
- `RPCHistory` is still only replacing the current behavior for `Subscription` method.

### Next steps:
- Expand `RPCHistory` usage to track all requests and responses, from local and remote origins, to provide a full history.
- Implement RPC responding for parsing and other errors on client side.
- Find a solution to avoid leaking cancellables on error calls.


Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -98,65 +110,69 @@ public final class RelayClient {
...
                 cancellable?.cancel()
                 onNetworkAcknowledge(error)
             }
         }
-        cancellable = requestAcknowledgePublisher
-            .filter {$0.id == request.id}
-            .sink { (_) in
-            cancellable?.cancel()
-                onNetworkAcknowledge(nil)
-        }
```
llbartekll, line 132: any advantage of changing the order?

Sources/WalletConnectUtils/RPCHistory.swift
```diff
@@ -0,0 +1,67 @@
+import JSONRPC
+
+public final class RPCHistory {
```
llbartekll, line 3: I wonder how new history can affect existing integrations 🤔
any thoughts?

Sources/WalletConnectRelay/RelayClient.swift
MisterVants: Setting up the events before the call prevents any race conditions in case of local socket errors. It's just a good practice

Sources/WalletConnectUtils/RPCHistory.swift
MisterVants: Currently, the history is being used just as a means of message deduplication, so the worst case scenario risk is the SDK receiving a duplicate `Subscription` once, which looks like a small trade-off to migrate to a history that can track all requests (the current implementation is generic, so it's bound to a single type of params). What do you think?

MisterVants: Also, we still have the transition to version 1.0.0 in the future as an opportunity to include all breaking changes that we need.

Sources/WalletConnectRelay/RelayClient.swift
llbartekll: Socket error will be handled in the completion block on a different thread so I think it's not the case.
Subscription will happen right after `send()` method execution as its just adds execution block.
correct me if I am wrong but imo the risk here is close to 0

if `send` would be a throwing function, error could happen before subscription but if error happens, you do not even care about subscription. anyway it does not apply for above code.

Sources/WalletConnectUtils/RPCHistory.swift
llbartekll: I think we want to avoid any breaking changes that could affect apps on production. But currently I do not think it will be very dangerous. Anyway it's worth to test it before merging.

Sources/WalletConnectRelay/RelayClient.swift
llbartekll: anyway, you can leave it like this, won't argue 😄

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -194,48 +218,38 @@ public final class RelayClient {
...
+        } else if let response = tryDecode(RPCResponse.self, from: payload) {
+            switch response.outcome {
+            case .success(let anyCodable):
+                if let _ = try? anyCodable.get(Bool.self) { // TODO: Handle success vs. error
+                    requestAcknowledgePublisherSubject.send(response.id)
+                } else if let subscriptionId = try? anyCodable.get(String.self) {
+                    subscriptionResponsePublisherSubject.send((response.id, subscriptionId))
+                }
+            case .failure(let rpcError):
+                logger.error("Received error message from iridium network, code: \(rpcError.code), message: \(rpcError.message)")
```
llbartekll, line 246: maybe we could use `relay network` instead of `iridium network`? as it is suppose to change.

Sources/WalletConnectSign/NetworkInteractor/NetworkRelaying.swift
```diff
@@ -11,9 +11,9 @@ protocol NetworkRelaying {
     func disconnect(closeCode: URLSessionWebSocketTask.CloseCode) throws
     func publish(topic: String, payload: String, tag: Int, prompt: Bool) async throws
     /// - returns: request id
-    @discardableResult func publish(topic: String, payload: String, tag: Int, prompt: Bool, onNetworkAcknowledge: @escaping ((Error?) -> Void)) -> Int64
+    func publish(topic: String, payload: String, tag: Int, prompt: Bool, onNetworkAcknowledge: @escaping ((Error?) -> Void))
```
llbartekll, line 14: 👍 yeah that was really not necessary to return an id.

llbartekll (commented):
Looks good! wish to see a generic networking interactor :)

Sources/WalletConnectRelay/RelayClient.swift
MisterVants: True, I've removed references to the "iridium" naming

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -171,20 +187,28 @@ public final class RelayClient {
         }
     }
 
-    @discardableResult public func unsubscribe(topic: String, completion: @escaping ((Error?) -> Void)) -> Int64? {
+    public func unsubscribe(topic: String, completion: @escaping ((Error?) -> Void)) {
         guard let subscriptionId = subscriptions[topic] else {
-            completion(RelyerError.subscriptionIdNotFound)
-            return nil
+            completion(RelayerError.subscriptionIdNotFound)
```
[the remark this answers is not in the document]
MisterVants: Renamed

Package.swift
```diff
@@ -36,7 +36,7 @@ let package = Package(
             path: "Sources/Auth"),
         .target(
             name: "WalletConnectRelay",
-            dependencies: ["WalletConnectUtils", "WalletConnectKMS"],
```
flypaper0, line 39: WalletConnectUtils depends on JSONRPC, do we need to add JSONRPC dependency here also?

Package.swift
```diff
@@ -71,11 +71,11 @@ let package = Package(
             dependencies: ["WalletConnectKMS", "WalletConnectUtils", "TestingUtils"]),
         .target(
             name: "TestingUtils",
-            dependencies: ["WalletConnectUtils", "WalletConnectKMS"],
+            dependencies: ["WalletConnectUtils", "WalletConnectKMS", "JSONRPC"],
             path: "Tests/TestingUtils"),
         .testTarget(
             name: "WalletConnectUtilsTests",
-            dependencies: ["WalletConnectUtils"]),
```
flypaper0, line 78: same here

Sources/WalletConnectRelay/RPC/RelayRPC.swift
```diff
@@ -0,0 +1,38 @@
...
+        return WalletConnectRPCID()
+    }
+
+    func wrapToIridium() -> PrefixDecorator<Self> {
+        return PrefixDecorator(rpcMethod: self, prefix: "iridium")
+    }
+
+    func wrapToIRN() -> PrefixDecorator<Self> {
+        return PrefixDecorator(rpcMethod: self, prefix: "irn")
+    }
```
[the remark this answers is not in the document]
flypaper0: I think we need to have a way to configure this in single place (not for every request). 

After next protocol renaming is better to replace method than add another `wrapToXXX` method 😁

Sources/Commons/Either.swift
```diff
@@ -63,3 +63,15 @@ extension Either: Codable where L: Codable, R: Codable {
         }
     }
 }
+
+extension Either: CustomStringConvertible {
```
flypaper0, line 67: Sorry, didn't catch. Why we need this `Either` abstraction?

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -98,65 +110,69 @@ public final class RelayClient {
 
     /// Completes when networking client sends a request, error if it fails on client side
     public func publish(topic: String, payload: String, tag: Int, prompt: Bool = false) async throws {
-        let params = RelayJSONRPC.PublishParams(topic: topic, message: payload, ttl: defaultTtl, prompt: prompt, tag: tag)
-        let request = JSONRPCRequest<RelayJSONRPC.PublishParams>(method: RelayJSONRPC.Method.publish.method, params: params)
-        logger.debug("Publishing Payload on Topic: \(topic)")
-        let requestJson = try request.json()
-        try await dispatcher.send(requestJson)
+        let request = Publish(params: .init(topic: topic, message: payload, ttl: defaultTtl, prompt: prompt, tag: tag))
+            .wrapToIridium()
```
flypaper0, line 114: It's easy to miss this wrap on new request implementation. We need to configure it once IG

Sources/Commons/Either.swift
MisterVants: It's used for the JSONRPC identifier, which, according to the specs, can be either a `String` or an `Int`

Package.swift
MisterVants: Removed it

Package.swift
MisterVants: Removed it

Sources/WalletConnectRelay/RelayClient.swift
MisterVants: In one of my first experiments, the wrap + transform into RPC needed to be called only in a single place before sending the message, but I couldn't fit it in the current implementation (because of the callback problems that I was trying to solve). I agree that it should be called just once.

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -98,65 +110,69 @@ public final class RelayClient {
...
-        let requestJson = try! request.json()
+        logger.debug("Relay: Subscribing to topic: \(topic)")
+        let rpc = Subscribe(params: .init(topic: topic))
+        let request = rpc
+            .wrapToIridium()
+            .asRPCRequest()
+        let message = try! request.asJSONEncodedString()
         var cancellable: AnyCancellable?
-        dispatcher.send(requestJson) { [weak self] error in
+        cancellable = subscriptionResponsePublisher
```
flypaper0, line 160: is it possible to replace completion with withCheckedThrowingContinuation? To make this method async?

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -98,65 +110,69 @@ public final class RelayClient {
...
+    ) {
+        let rpc = Publish(params: .init(topic: topic, message: payload, ttl: defaultTtl, prompt: prompt, tag: tag))
+        let request = rpc
+            .wrapToIridium()
+            .asRPCRequest()
+        let message = try! request.asJSONEncodedString()
+        logger.debug("Publishing Payload on Topic: \(topic)")
         var cancellable: AnyCancellable?
-        dispatcher.send(requestJson) { [weak self] error in
+        cancellable = requestAcknowledgePublisher
```
flypaper0, line 136: Need to find way how to move requestAcknowledgePublisher subscription out of this method.

Sources/WalletConnectRelay/RelayClient.swift
MisterVants: The problem is that we need to wait for the acknowledgement with the subscription ID, the same problem that I was talking about in out sync meeting, that I couldn't solve yet

Sources/WalletConnectRelay/RelayClient.swift
MisterVants: Yes, the RelayClient refactor is not over

Sources/WalletConnectRelay/RelayClient.swift
jjgikufg: merhaba birşey sormak istiyorum

flypaper0 (approved):

Sources/WalletConnectUtils/RPCHistory.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    public struct Record: Codable {
+        public enum Origin: String, Codable {
+            case local
+            case remote
+        }
+        let id: RPCID
+        let topic: String
+        let origin: Origin
```
llbartekll, line 12: just noticed it. What is the purpose of origin?

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -194,48 +218,38 @@ public final class RelayClient {
...
-            do {
-                try jsonRpcSubscriptionsHistory.set(topic: request.params.data.topic, request: request)
-                onMessage?(request.params.data.topic, request.params.data.message)
-                acknowledgeSubscription(requestId: request.id)
-            } catch {
-                logger.info("Relay Client Info: Json Rpc Duplicate Detected")
+        if let request = tryDecode(RPCRequest.self, from: payload) {
+            if let params = try? request.params?.get(Subscription.Params.self) {
+                do {
+                    try rpcHistory.set(request, forTopic: params.data.topic, emmitedBy: .remote)
```
llbartekll, line 228: it is the only place where we are setting a history record. Should we record history when requesting as well?

Sources/WalletConnectUtils/RPCHistory.swift
MisterVants: Being able to track whether a request was sent from the client or from the peer. Thinking on long term support, I think is a nice thing to have, might be useful in the future

Sources/WalletConnectRelay/RelayClient.swift
MisterVants: Right now, it's the only place. I'm in for recording all requests, so we can have a complete history.

PR #385, opened 2022-07-29
#376 auth request service

llbartekll:
- implements authRequestService
- adds auth granular tag
- implements networking interactor's request method


flypaper0 (approved):

Sources/Auth/Types/AuthRequestParams.swift
```diff
@@ -0,0 +1,33 @@
...
+}
+
+extension AuthRequestParams {
+    struct Requester: Codable, Equatable {
+        let publicKey: String
+        let metadata: AppMetadata
+    }
+}
+
+// TODO - temporarly duplicated - moved do utils in concurrent PR
```
MisterVants, line 21: I believe that, if this is moved to Utils, we'll have to expose it via a `typealias` in each SDK. Remember to test it if you do

Sources/Auth/Services/App/AuthRequestService.swift
```diff
@@ -1,8 +1,30 @@
...
+    }
 
+    func request(params: RequestParams, topic: String) async throws {
+        let pubKey = try kms.createX25519KeyPair()
+        let responseTopic = pubKey.rawRepresentation.sha256().toHexString()
+        let requester = AuthRequestParams.Requester(publicKey: pubKey.hexRepresentation, metadata: appMetadata)
+        let issueAt = ISO8601DateFormatter().string(from: Date())
+        let payload = AuthPayload(requestParams: params, iat: issueAt)
+        let params = AuthRequestParams(requester: requester, payloadParams: payload)
+        let request = JSONRPCRequest<AuthRequestParams>(method: "wc_authRequest", params: params)
```
MisterVants, line 25: What do you think of using `RPCRequest` in here? Seems like a good place to use it

Sources/Auth/Services/Common/NetworkingInteractor.swift
```diff
@@ -1,18 +1,40 @@
...
+    func request(_ request: JSONRPCRequest<AuthRequestParams>, topic: String, envelopeType: Envelope.EnvelopeType = .type0) async throws {
+        try await self.request(request, topic: topic, envelopeType: envelopeType)
+    }
+}
+
+
 class NetworkingInteractor: NetworkInteracting {
     private let relayClient: RelayClient
+    private let serializer: Serializing
+    private let jsonRpcHistory: JsonRpcHistory<AuthRequestParams>
```
MisterVants, line 21: As it was already merged to develop, what do you also think of using the improved `RPCHistory` in here, so you can avoid binding the storage to a single specialized type?

MisterVants:
Left some suggestions to incentivise scaling the service from the start. Also, the PR needs to run again in the CI (maybe rebase with develop)

Sources/Auth/Services/App/AuthRequestService.swift
llbartekll: I was waiting for your first demo integration 😄  but ok, let's give it a shot.

PR #397, opened 2022-08-02
[Auth] #395 Auth Request Subscriber

### LINKED ISSUE #395: Auth Request Subscriber
Auth Request Subscriber subscribes for `wc_authRequest` requests and delegates Auth Client with formatted `eip-4361` message.

llbartekll:
close #395


Sources/Auth/Services/Wallet/AuthRequestSubscriber.swift
```diff
@@ -0,0 +1,37 @@
...
+    private func subscribeForRequest() {
+        networkingInteractor.requestPublisher.sink { [unowned self] subscriptionPayload in
+            guard subscriptionPayload.request.method == "wc_authRequest" else { return }
+            guard let authRequest = try? subscriptionPayload.request.params?.get(AuthRequestParams.self) else {
+                logger.debug("Malformed auth request params")
+                return
+            }
+            do {
+                let message = try messageFormatter.formatMessage(from: authRequest)
+                guard let requestId = subscriptionPayload.request.id?.right else { return }
```
llbartekll, line 30: @MisterVants I began to integrate you rpc lib and I do not like this part of querying id from request. Do you guys think we need `Either` type for request id? It is totally unexpected to have a String type request id in the network.

@flypaper0 ⬆️

flypaper0: I doesn't know do we really need Either or not, but we can make it CustomStringConvertible to work with `id` like with regular String

flypaper0: But may be better to make id - String, but make possibility to parse Int/String id param from RPC response

MisterVants: You should not query the request ID internal value as an Int or String. The `RPCID` type already abstracts that and should be treated as a primitive. Just use it as you would use an integer ID:

`var onRequest: ((_ id: RPCID, _ message: String)->())?`

`guard let requestId = subscriptionPayload.request.id else { ... }`

MisterVants: (Might need to `import JSONRPC`)

Sources/Auth/Services/Wallet/AuthRequestSubscriber.swift
```diff
@@ -0,0 +1,37 @@
+import Combine
+import Foundation
+import WalletConnectUtils
+
+class AuthRequestSubscriber {
+    private let networkingInteractor: NetworkInteracting
+    private let logger: ConsoleLogging
+    private var publishers = [AnyCancellable]()
+    private let messageFormatter: SIWEMessageFormatting
+    var onRequest: ((_ id: Int64, _ message: String)->())?
```
MisterVants, line 10:
```suggestion
    var onRequest: ((_ id: RPCID, _ message: String)->())?
```

Sources/Auth/Services/Wallet/AuthRequestSubscriber.swift
MisterVants, line 30:
```suggestion
                guard let requestId = subscriptionPayload.request.id else { return }
```

Sources/JSONRPC/RPCID.swift
```diff
@@ -1,6 +1,6 @@
 import Commons
 
-public typealias RPCID = Either<String, Int>
+public typealias RPCID = Either<String, Int64>
```
MisterVants, line 3: I had a question when implementing this: Is there any possibility that the SDK is going to run in a 32-bit system?

Sources/Auth/Services/Wallet/AuthRequestSubscriber.swift
llbartekll: I am not sure if it is a good idea to make all classes passing RPCID up to the app and make them all dependent to `JSONRPC`. 
Another inconvenience is that id is an optional

llbartekll:
@MisterVants committed your RPCID suggestion, maybe later we can use typealias to not import JSONRPC lib all over the sdk

Sources/Auth/Services/Wallet/AuthRequestSubscriber.swift
MisterVants: The typealias you suggested might be a good idea to avoid imports, and helps decoupling in case of any future changes. It's a primitive, after all.

The optional ID is required by the jsonrpc 2.0 specs to expect notification requests. Unfortunately, this is essential to make the RPC layers future-proof

MisterVants (approved):

PR #399, opened 2022-08-02
[Refactor] #391 Remove unused code in network interacting

MisterVants:
Self-explanatory

### DIFF
--- a/Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
+++ b/Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
@@ -140,7 +140,7 @@ final class ApproveEngine {
         try await networkingInteractor.subscribe(topic: topic)
         sessionStore.setSession(session)
 
-        try await networkingInteractor.request(.wcSessionSettle(settleParams), onTopic: topic)
+        networkingInteractor.request(.wcSessionSettle(settleParams), onTopic: topic)
         onSessionSettle?(session.publicRepresentation())
     }
 }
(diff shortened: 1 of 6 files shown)


Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
```diff
@@ -140,7 +140,7 @@ final class ApproveEngine {
         try await networkingInteractor.subscribe(topic: topic)
         sessionStore.setSession(session)
 
-        try await networkingInteractor.request(.wcSessionSettle(settleParams), onTopic: topic)
+        networkingInteractor.request(.wcSessionSettle(settleParams), onTopic: topic)
```
flypaper0, line 143: Why async call removed?

Sources/WalletConnectSign/NetworkInteractor/NetworkInteractor.swift
```diff
@@ -62,18 +59,6 @@ class NetworkInteractor: NetworkInteracting {
         setUpPublishers()
     }
 
-    func request(_ wcMethod: WCMethod, onTopic topic: String) async throws {
```
flypaper0, line 65: the idea was replace `request` method with async `request`. But here you removed all async calls

MisterVants:
Removing the async method makes it impossible to handle errors beyond the interactor call. Changing it to draft.

Sources/WalletConnectSign/NetworkInteractor/NetworkInteractor.swift
llbartekll: totally agree, do not get the purpose of doing so.

llbartekll (commented):
Can we close it?

MisterVants:
> Can we close it?

yes

PR #401, opened 2022-08-03
[Auth] #377 Auth respond service 

llbartekll:


Sources/JSONRPC/RPCResponse.swift
```diff
@@ -35,23 +35,27 @@ public struct RPCResponse: Equatable {
...
     }
 
     public init<C>(id: String, result: C) where C: Codable {
         self.init(id: RPCID(id), outcome: .success(AnyCodable(result)))
     }
 
-    public init(id: Int, error: JSONRPCError) {
+    public init<C>(id: RPCID, result: C) where C: Codable {
+        self.init(id: id, outcome: .success(AnyCodable(result)))
+    }
```
MisterVants, line 48: I see that you needed an init that receives a `RPCID`. In your opinion, do you think it's better to add these specific inits or should we change the default init from `internal` to `public`?

MisterVants (approved):

Issue #364, opened 2022-07-25
Support `irn` instead of `iridium`


pedrouid:
Related to https://github.com/WalletConnect/walletconnect-specs/issues/90

llbartekll:
@MisterVants can we close this one? is it included in your last PR?

flypaper0:
> @MisterVants can we close this one? is it included in your last PR?

there is implemented "inr" "iridium" wrappers, but "iridium" wrapper still in use

MisterVants:
I'm replacing the calls

PR #402, opened 2022-08-03
[Network] #364: Change relay methods prefix from "iridium" to "irn"

MisterVants:
#closes 364

Self-explanatory

### DIFF
--- a/Sources/WalletConnectPairing/Types/WCPairing.swift
+++ b/Sources/WalletConnectPairing/Types/WCPairing.swift
@@ -37,7 +37,7 @@ public struct WCPairing: SequenceObject {
 
     public init(topic: String) {
         self.topic = topic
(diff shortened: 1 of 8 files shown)


flypaper0 (approved):

llbartekll:
let's change relay protocol options for irn as well. so uri do not contain legacy naming

llbartekll (changes requested):
let's change relay protocol options for irn as well. so uri do not contain legacy naming

llbartekll (commented):
I am afraid that changing storage identifier will break existing integrations. Can you manually test it?

MisterVants:
I think we can just leave as it is, it's well isolated and won't affect the protocol

llbartekll (approved):

PR #405, opened 2022-08-04
[Auth] Cacao signer

flypaper0:
**What changed**

Cacao object EIP-191 signature generation and verification


Tests/AuthTests/CacaoSignerTests.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+import XCTest
+@testable import Auth
+import TestingUtils
+
+class CacaoSignerTest: XCTestCase {
+
+    let payload = CacaoPayload(
+        iss: "did:pkh:eip155:1:0x22Fe071b3631f155F0d8f4c9377D3309cB904E10",
```
llbartekll, line 9: Maybe we could create a DIDPKH object co we can easily derive caip-10 account or an address from it?

flypaper0: Check SignerAddress please. There is `static func from(iss: String) -> String`
Looks good for you?

llbartekll: I was thinking more about something like:
 ```
let signerDid = DIDPKH(account: Account)
signerDid.account ...
```
but if it works it works.

File: Sources/Auth/Services/Signer/CocoaSigner.swift [hunk 2]
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+protocol CacaoSignatureVerifying {
+    func verifySignature(_ cacao: Cacao) throws
+}
+
+struct CacaoSigner: CacaoSignatureVerifying {
```
MisterVants, line 7: The file name is `CocoaSigner` while the type is `CacaoSigner`

Sources/Auth/Services/Signer/CocoaSigner.swift
```diff
@@ -0,0 +1,32 @@
...
+        case signatureValidationFailed
+    }
+
+    private let signer: Signer
+
+    init(signer: Signer) {
+        self.signer = signer
+    }
+
+    func sign(payload: CacaoPayload, privateKey: Data) throws -> CacaoSignature {
```
MisterVants, line 19: Passing private keys around in the SDK products' layer worries me that we might be handling the keys too loosely. Can we think a way of dealing with private keys only inside the KMS module?

MisterVants (approved):

Sources/Auth/Services/Signer/CocoaSigner.swift
flypaper0: Good point but client of `sign` method unimplemented yet. I suggest to think about this when it will be done

Sources/Auth/Services/Signer/CocoaSigner.swift
flypaper0: thanks

Sources/Auth/Services/Signer/CacaoSigner.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    func sign(payload: CacaoPayload, privateKey: Data) throws -> CacaoSignature {
+        let message = try JSONEncoder().encode(payload) // TODO: SIWE encoding
+        let signature = try signer.sign(message: message, with: privateKey)
+        return CacaoSignature(t: "eip191", s: signature.toHexString(), m: String())
+    }
+
+    func verifySignature(_ cacao: Cacao) throws {
+        let sig = Data(hex: cacao.signature.s)
+        let message = try JSONEncoder().encode(cacao.payload)
```
llbartekll, line 27: would it maybe better to add a new method to `SIWEMessageFormatting`
`formatMessage(from cacao: Cacao) -> String`?
for the implementation we can have a separate issue but we are not suppose to verify signature on cacao but on EIP-4361 message instead.

Sources/Auth/Services/Signer/CacaoSigner.swift
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    private let signer: Signer
+
+    init(signer: Signer) {
+        self.signer = signer
+    }
+
+    func sign(payload: CacaoPayload, privateKey: Data) throws -> CacaoSignature {
+        let message = try JSONEncoder().encode(payload) // TODO: SIWE encoding
```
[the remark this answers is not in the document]
llbartekll: and then I think it is just utf8 encoding into data

Package.swift
```diff
@@ -17,10 +17,12 @@ let package = Package(
             name: "WalletConnectChat",
             targets: ["Chat"]),
         .library(
-            name: "WalletConnectPairing",
-            targets: ["WalletConnectPairing"])
+            name: "WalletConnectAuth",
+            targets: ["Auth"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/flypaper0/Web3.swift", .branch("master"))
```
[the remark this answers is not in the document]
llbartekll: maybe we can open issue to not forget of changing this...

Tests/AuthTests/CacaoSignerTests.swift
```diff
@@ -0,0 +1,50 @@
...
+        iat: "2022-03-10T17:09:21.481+03:00",
+        nbf: "2022-03-10T17:09:21.481+03:00",
+        exp: "2022-03-10T18:09:21.481+03:00",
+        statement: "I accept the ServiceOrg Terms of Service: https://service.org/tos",
+        requestId: "request-id-random",
+        resources: [
+            "ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq",
+            "https://example.com/my-web2-claim.json"
+        ]
+    )
```
llbartekll, line 25: I think we should test on EIP-4361 message

llbartekll (changes requested):

Package.swift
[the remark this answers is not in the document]
flypaper0: https://github.com/WalletConnect/WalletConnectSwiftV2/issues/426

Sources/Auth/Services/App/AuthRespondSubscriber.swift
```diff
@@ -7,30 +7,48 @@ class AuthRespondSubscriber {
...
+
             do {
-                try CacaoSignatureVerifier().verifySignature(cacao)
-                onResponse?(subscriptionPayload.request.id!, cacao)
+                guard let cacao = try subscriptionPayload.request.result?.get(Cacao.self) else {
+                    return logger.debug("Malformed auth response params")
+                }
+
+                let address = try DIDPKH(iss: cacao.payload.iss).account.address
+                let payload = AuthPayload(payload: cacao.payload)
```
llbartekll, line 43: instead of initialising `AuthPayload` from cacao I would rather add new method to messageFormatter:
`messageFormatter.formatMessage(from: cacao)`

Sources/Auth/Services/App/AuthRespondSubscriber.swift
```diff
@@ -7,30 +7,48 @@ class AuthRespondSubscriber {
...
+                let message = messageFormatter.formatMessage(from: payload, address: address)
+
+                try signatureVerifier.verify(
+                    signature: cacao.signature.s,
+                    message: message,
+                    address: cacao.payload.iss
+                )
```
llbartekll, line 50: here you know if signature is valid or not but you do not know if signed object was not malformed.
I think you should compare message constructed from requested (authpayload + issuer) against message from cacao.
and if they are not the same it means that wallet maliciously modified some properties.

Sources/Auth/Services/Signer/MessageSigner.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+
+protocol MessageSignatureVerifying {
+    func verify(signature: String, message: String, address: String) throws
+}
+
+protocol MessageSignatureSigning {
```
llbartekll, line 7: I do not like `SignatureSigning ` naming 😄

Sources/Auth/Services/Signer/MessageSigner.swift
```diff
@@ -0,0 +1,36 @@
...
+    func sign(message: String, privateKey: Data) throws -> String
+}
+
+struct MessageSigner: MessageSignatureVerifying & MessageSignatureSigning {
```
llbartekll, line 11: Would make sense to split it into two separate classes?
SIWESigner and SIWEVerifier?
and maybe throw an error if message argument in function call is not a valid SIWE message?

Sources/Auth/Services/App/AuthRespondSubscriber.swift
llbartekll: correct me if I am wrong but I think this is a important security issue

Sources/Auth/Types/AuthPayload.swift
```diff
@@ -28,4 +28,19 @@ struct AuthPayload: Codable, Equatable {
...
+        self.aud = payload.aud
+        self.version = payload.version
+        self.nonce = payload.nonce
+        self.iat = payload.iat
+        self.nbf = payload.nbf
+        self.exp = payload.exp
+        self.statement = payload.statement
+        self.requestId = payload.requestId
+        self.resources = payload.resources
+    }
```
llbartekll, line 45: I would use this object for networking requests only.
we can actually move it into `ProtocolRPCParams` folder

llbartekll (changes requested):

Sources/Auth/Services/Signer/MessageSigner.swift
flypaper0: This is not a real SIWESigner/Verifier. It will for property for regular strings too. I think need to live it in a single class. Splitted protocols just encapsulate methods usage in app/wallet parts

Sources/Auth/Types/AuthPayload.swift
flypaper0: removed this initializer

llbartekll (approved):

Issue #394, opened 2022-08-02
SIWE Message Formatter


llbartekll:
SIWE Message Formatter is expected to format `AuthRequestParams` into  ABNF Message Formatted EIP4361 payload 
https://eips.ethereum.org/EIPS/eip-4361
```
struct SIWEMessageFormatter {
    func formatMessage(from request: AuthRequestParams) throws -> String {}
}
```

PR #406, opened 2022-08-05
[Auth] #394 SIWE messge formatter

llbartekll:
add SIWE message formatter

### DIFF
--- a/Sources/Auth/Services/Common/SIWEMessageFormatter.swift
+++ b/Sources/Auth/Services/Common/SIWEMessageFormatter.swift
@@ -9,3 +9,55 @@ struct SIWEMessageFormatter: SIWEMessageFormatting {
         fatalError("not implemented")
     }
 }
+
+struct SIWEMessage: Equatable {
+    let domain: String
+    let uri: String //aud
+    let address: String
+    let version: String
+    let nonce: String
+    let chainId: String
+    let type: String
+    let iat: String
+    let nbf: String?
+    let exp: String?
+    let statement: String?
+    let requestId: String?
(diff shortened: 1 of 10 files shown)


Sources/Auth/Services/Common/SIWEMessageFormatter.swift
```diff
@@ -1,11 +1,91 @@
...
+                URI: \(uri)
+                Version: \(version)
+                Chain ID: \(chainId)
+                Nonce: \(nonce)
+                Issued At: \(iat)\(expLine)\(nbfLine)\(requestIdLine)\(resourcesSection)
+                """
+    }
+
+    var expLine: String {
+        if let exp = exp {
```
flypaper0, line 54: may be guard? Here and in other params

Sources/Auth/Services/Common/SIWEMessageFormatter.swift
```diff
@@ -1,11 +1,91 @@
...
+            return "\nRequest ID: \(requestId)"
+        }
+        return ""
+    }
+
+    var resourcesSection: String {
+        var section = ""
+        if let resources = resources {
+            section = "\nResources:"
+            resources.forEach {
```
flypaper0, line 85: good to use reduce here or join(separator: )

llbartekll: ok it is shorter

flypaper0 (changes requested):
need to make work testNilOptionalParamsMessage test

flypaper0 (approved):

PR #412, opened 2022-08-08
v0.9.3-rc.0

llbartekll:


flypaper0 (approved):

MisterVants (changes requested):
I believe the release PR is lacking the version bump in user agent.

llbartekll:
@MisterVants good catch. how we can automate it ?🤔

MisterVants (approved):

MisterVants:
> @MisterVants good catch. how we can automate it ?🤔

Good question, I don't know. Is it possible to make a CI check that tries to match the release name with the user agent info? That could raise an error on the PR, at least.

PR #421, opened 2022-08-09
#382 auth networking interactor

llbartekll:
Add `requestNetworkAck` function that completes with an acknowledgement from the relay network.
handles incoming requests and responses to networking interactor and publishes decrypted rpc requests to subscribers

### DIFF
--- a/Sources/Auth/Services/Common/NetworkingInteractor.swift
+++ b/Sources/Auth/Services/Common/NetworkingInteractor.swift
@@ -73,10 +73,11 @@ class NetworkingInteractor: NetworkInteracting {
             let message = try serializer.serialize(topic: topic, encodable: request)
             return try await withCheckedThrowingContinuation { continuation in
                 relayClient.publish(topic: topic, payload: message, tag: tag) { error in
-                    guard let error1 = error else {
-                        fatalError("Expected non-nil result 'error1' in the non-error case")
+                    if let error = error {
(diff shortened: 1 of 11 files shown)


flypaper0 (approved):
lgtm, little task descriptions in PR's can be helpful

Sources/Auth/Services/Common/NetworkingInteractor.swift
```diff
@@ -11,6 +11,7 @@ protocol NetworkInteracting {
     func subscribe(topic: String) async throws
     func unsubscribe(topic: String)
     func request(_ request: RPCRequest, topic: String, tag: Int, envelopeType: Envelope.EnvelopeType) async throws
+    func requestNetworkAck(_ request: RPCRequest, topic: String, tag: Int) async throws
```
MisterVants, line 14: We need to get rid of this kind of function that waits for the network acknowledgement (at least if we want to generalize the interactor). We need to find a better way around to wait for the acknowledgement on calls that need it.

MisterVants (approved):

Sources/Auth/Services/Common/NetworkingInteractor.swift
llbartekll: I am not sure I understand, what you mean by `generalize`? isn't this approach generic and interface could be shared by all SDKs?

PR #423, opened 2022-08-09
Rename target: `TestingUtils` to `Toolbox`

MisterVants:
Renames the target, as talked some months ago, to better convey identity.


MisterVants:
@flypaper0 same issue here of CI not resolving the package graph, do you know how to solve?

flypaper0:
> @flypaper0 same issue here of CI not resolving the package graph, do you know how to solve?

Check this commit in CacaoSigner PR https://github.com/WalletConnect/WalletConnectSwiftV2/pull/405/commits/b1f3b5ed5fb20248786d81c31060377662cc3f6f. But in your case another run could help

llbartekll (commented):
I am not sure I like it more

PR #424, opened 2022-08-09
[Core] WalletConnectRouter

### LINKED ISSUE #233: Switch back to dapp
**Is your feature request related to a problem? Please describe.**

Support a way for wallet apps to switch back to the client dapp running the browser when both the wallet and browser are on the same mobile device

**Describe the solution you'd like**
1. Visit dapp in iOS Safari
2. Choose WC
3. Choose Wallet
4. Auto switched to Wallet
5. Accept connection in Wallet.

**Additional context**

Feature request from AlphaWallet.

Metamask and Rainbow have done something like this.

They trigger a native API to switch back to previous app in foreground It seems to be a routing API that you can go back in history once.

Here is the source code that shows how Rainbow use “goBack” in their package
https://github.com/brunobar79/react-native-minimizer/blob/master/ios/Minimizer.m

flypaper0:
close #233

**What changed**

WalletConnectRouter wrap `goBack()` action for external usage. Usage example added in sample app

**Sample app**

https://user-images.githubusercontent.com/5384197/183735032-52394713-e7ab-4981-ae64-4171151a7ac6.mov

### DIFF
--- a/Example/DApp/SelectChain/SelectChainView.swift
+++ b/Example/DApp/SelectChain/SelectChainView.swift
@@ -2,16 +2,27 @@ import Foundation
 import UIKit
 
 class SelectChainView: UIView {
+
     let tableView: UITableView = {
         let tableView = UITableView(frame: .zero, style: .insetGrouped)
         tableView.backgroundColor = .tertiarySystemBackground
         tableView.register(UITableViewCell.self, forCellReuseIdentifier: "chain")
         return tableView
     }()
+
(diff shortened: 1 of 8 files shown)


llbartekll (approved):
Good Work! probably we should document it.

chadyj:
Please also update the mobile linking docs here when this is merged https://github.com/WalletConnect/walletconnect-docs/pull/234

PR #425, opened 2022-08-10
[Auth] #378 auth client

llbartekll:
- Implements Auth Client's methods
- add cleanup service for debugging


Sources/Auth/AuthClient.swift
```diff
@@ -1,31 +1,121 @@
...
+
+    public func respond(_ params: RespondParams) async throws {
+        guard let account = account else { throw Errors.unknownWalletAddress }
+        try await walletRespondService.respond(respondParams: params, account: account)
+    }
+
+    public func getPendingRequests() throws -> [AuthRequest] {
+        guard let account = account else { throw Errors.unknownWalletAddress }
+        let pendingRequests: [AuthRequest] = rpcHistory.getPending()
+            .filter {$0.request.method == "wc_authRequest"}
```
flypaper0, line 94: mb move filter and mapping out of AuthClient? rpcHistoryService or pendingService

Sources/Auth/AuthClient.swift
```diff
@@ -1,31 +1,121 @@
...
 
-    func pair(uri: String) async throws {
+    public func pair(uri: String) async throws {
         guard let pairingURI = WalletConnectURI(string: uri) else {
             throw Errors.malformedPairingURI
         }
         try await walletPairService.pair(pairingURI)
     }
+
+    public func request(_ params: RequestParams, topic: String?) async throws -> String? {
```
flypaper0, line 70: let's split it into two methods?

Sources/Auth/AuthClient.swift
llbartekll: good idea!

Sources/Auth/AuthClient.swift
llbartekll: 🤔

flypaper0 (approved):

Sources/Auth/AuthClient.swift
llbartekll: actually I agree, looks better

PR #436, opened 2022-08-12
[Hotfix] EIP-155 signature fix

flypaper0:
***What changed***
Replay protection implemented


MisterVants (approved):

llbartekll (approved):
I thought nonce is the only reply attack protection.
missed it specs

PR #437, opened 2022-08-14
Adds more unit tests for WCPairing

MaisaMilena:
- Rename some tests to specify if is testing init from uri or topic 
- Add scenarios that throw error 

While working on this I noticed that the `activate` function sets the status to `true` even if `updateExpiry` throws an [error](https://github.com/WalletConnect/WalletConnectSwiftV2/blob/b676369f6dd72d8d0d80fe1275313f1f74cade7b/Sources/WalletConnectPairing/Types/WCPairing.swift#L52). Is that the expected behavior? Maybe `activate` should only change the `active` var?

### DIFF
--- a/Tests/WalletConnectSignTests/WCPairingTests.swift
+++ b/Tests/WalletConnectSignTests/WCPairingTests.swift
@@ -75,21 +75,27 @@ final class WCPairingTests: XCTestCase {
     }
     
     func testUpdateExpiry_WhenNewExpiryDateIsLessThanExpiryDate_ShouldThrowInvalidUpdateExpiryValue() {
-        let expiryDate = referenceDate.advanced(by: 10 * .minute)
+        let expiryDate = referenceDate.advanced(by: 40 * .day)
         var pairing = WCPairing(topic: "", relay: .stub(), peerMetadata: .stub(), expiryDate: expiryDate)
-        XCTAssertThrowsError(try pairing.updateExpiry(40 * .day)) { error in
+        XCTAssertThrowsError(try pairing.updateExpiry(10 * .minute)) { error in
             XCTAssertEqual(error as! WCPairing.Errors, WCPairing.Errors.invalidUpdateExpiryValue)
         }
     }
 
-    func testActivate_WhenUpdateExpiryFails_ShouldActivateAndThrowError() {
-        let expiryDate = referenceDate.advanced(by: 10 * .minute)
+    func testActivate_WhenCanUpdateExpiry_ShouldActivateAndUpdateExpiryIn30Days() {
+        var pairing = WCPairing(topic: "", relay: .stub(), peerMetadata: .stub(), expiryDate: referenceDate)
+        XCTAssertFalse(pairing.active)
+        pairing.activate()
+        XCTAssertTrue(pairing.active)
+        XCTAssertEqual(referenceDate.advanced(by: 30 * .day), pairing.expiryDate)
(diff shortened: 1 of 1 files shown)


MaisaMilena:
Btw, while on `development` running the tests, `CacaoSignerTest` and `SignerTest` failed

MisterVants (changes requested):
Test needs a fix (update less than expiry)

Tests/WalletConnectSignTests/WCPairingTests.swift
```diff
@@ -35,18 +35,61 @@ final class WCPairingTests: XCTestCase {
...
+            XCTAssertEqual(error as! WCPairing.Errors, WCPairing.Errors.invalidUpdateExpiryValue)
+        }
+    }
+
+    func testActivate_WhenUpdateExpiryFails_ShouldActivateAndThrowError() {
+        let expiryDate = referenceDate.advanced(by: 10 * .minute)
+        var pairing = WCPairing(topic: "", relay: .stub(), peerMetadata: .stub(), expiryDate: expiryDate)
+        XCTAssertFalse(pairing.active)
+        pairing.activate()
+        XCTAssertThrowsError(try pairing.updateExpiry(10 * .minute)) { error in
```
MisterVants, line 90: `activate()` makes a call to `updateExpiry()` internally (which sets a 30-day expiry). 

Is there a reason to call activate before a throwing update? Does it make a difference for this test to invert the order?

Tests/WalletConnectSignTests/WCPairingTests.swift
```diff
@@ -35,18 +35,61 @@ final class WCPairingTests: XCTestCase {
...
+        var pairing = WCPairing(topic: "", relay: .stub(), peerMetadata: .stub(), expiryDate: referenceDate)
+        XCTAssertThrowsError(try pairing.updateExpiry(40 * .day)) { error in
+            XCTAssertEqual(error as! WCPairing.Errors, WCPairing.Errors.invalidUpdateExpiryValue)
+        }
+    }
+    
+    func testUpdateExpiry_WhenNewExpiryDateIsLessThanExpiryDate_ShouldThrowInvalidUpdateExpiryValue() {
+        let expiryDate = referenceDate.advanced(by: 10 * .minute)
+        var pairing = WCPairing(topic: "", relay: .stub(), peerMetadata: .stub(), expiryDate: expiryDate)
+        XCTAssertThrowsError(try pairing.updateExpiry(40 * .day)) { error in
```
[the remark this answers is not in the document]
MaisaMilena: ops! 😅

Tests/WalletConnectSignTests/WCPairingTests.swift
MaisaMilena: I was trying to check the error inside the `activate()` function, but if I understood it right the error dies inside `activate()`. I changed the test, now checking for the `expiryDate` update. If an error occurs, it shouldn't be updated.

flypaper0:
lgtm, more tests is always better 👍
fix testcase naming pls "testUpdateExpiry_WhenValueIsGreaterThanMax_ShouldThrowInvalidUpdateExpiryValue"
Little less text and without "_"

flypaper0 (approved):

MaisaMilena:
@flypaper0 updated! Do these new names fit better?

MisterVants (approved):

PR #438, opened 2022-08-15
[Relay] Relay Singleton instance

### LINKED ISSUE #430: Replace Relays client's `onMessage` callback with combine publisher in order to allow 1 to many relation

flypaper0:
# Description

* Relay singleton wrapper
* Replace onMessage with combine publisher
* SignClient Delegate removed, Sign singleton is SignClient instance
* RelayClient onMessage replaced with combine publisher

Resolves #430 #431 #432

## How Has This Been Tested?

* run unit, integration tests

## Due Dilligence

* [ ] Breaking change
* [x] Requires a documentation update (@llbartekll Sign Interface doesn't changed, should we update docs for this?)
https://github.com/WalletConnect/walletconnect-docs/pull/236


llbartekll (commented):
Good work! much cleaner.
I just wonder if we should not require relay configuration separately, consider that wallet want to integrate sign and chat alongside wouldn't it be better if configuration looks as follows?:
1. relay.configure(projectId etc.)
2. chat.configure()
3. sign.configure()
imo currently is better for only one sdk because we do not expose relay configuration but it will e quite weird if we require:
1 . chat.configure(projectId? etc)
2 . sign.configure(projectId? etc)

MisterVants (approved):

Example/IntegrationTests/Sign/Helpers/ClientDelegate.swift
```diff
@@ -1,10 +1,8 @@
 import Foundation
 @testable import WalletConnectSign
+import Combine
 
-class ClientDelegate: SignClientDelegate {
-    func didChangeSocketConnectionStatus(_ status: SocketConnectionStatus) {
-        onConnected?()
-    }
+class ClientDelegate {
```
llbartekll, line 5: actually when we do not depend on sign delegate anymore we could remove this class entirely.
maybe we can create an issue and do it at some point?

llbartekll (approved):

Example/IntegrationTests/Sign/Helpers/ClientDelegate.swift
flypaper0: ClientDelegate usage looks good in tests for me

llbartekll: we could subscribe for SignClients events only now

PR #449, opened 2022-08-18
[Auth] Handle errors

### LINKED ISSUE #445: Handle error auth response

flypaper0:
# Description
- App / Wallet subscribers error handling
- InternalError and ExternalError implemented
- WalletRespondService respond with error

Resolves #445

## How Has This Been Tested?
Run integration and Auth unit tests

## Due Dilligence

* [ ] Breaking change


Example/IntegrationTests/Auth/AuthTests.swift
```diff
@@ -79,6 +79,6 @@ final class AuthTests: XCTestCase {
             responseExpectation.fulfill()
         }
         .store(in: &publishers)
-        wait(for: [responseExpectation], timeout: 2)
+        wait(for: [responseExpectation], timeout: .infinity)
```
llbartekll, line 82: maybe 5, 10 seconds?

Sources/Auth/AuthClient.swift
```diff
@@ -15,8 +15,8 @@ public class AuthClient {
         authRequestPublisherSubject.eraseToAnyPublisher()
     }
 
-    private var authResponsePublisherSubject = PassthroughSubject<(id: RPCID, result: Result<Cacao, ErrorCode>), Never>()
-    public var authResponsePublisher: AnyPublisher<(id: RPCID, result: Result<Cacao, ErrorCode>), Never> {
+    private var authResponsePublisherSubject = PassthroughSubject<(id: RPCID, result: Result<Cacao, InternalError>), Never>()
```
llbartekll, line 18: I think the error can be a protocol error.
consider case when wallet rejects a request - user initiated

Sources/Auth/AuthClient.swift
```diff
@@ -87,9 +87,9 @@ public class AuthClient {
         try await appRequestService.request(params: params, topic: topic)
     }
 
-    public func respond(_ result: Result<RespondParams, ErrorCode>) async throws {
+    public func respond(requestId: RPCID, result: Result<CacaoSignature, ExternalError>) async throws {
```
llbartekll, line 90: I think here it should be a protocol error as well as it is send to a peer

Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
```diff
@@ -2,48 +2,58 @@ import Combine
...
+            guard let authRequestParams = try? payload.request.params?.get(AuthRequestParams.self)
+            else { return respondError(.malformedRequestParams, topic: payload.topic, requestId: requestId) }
+
+            let message = messageFormatter.formatMessage(from: authRequestParams.payloadParams, address: address)
+
+            onRequest?(requestId, message)
         }.store(in: &publishers)
     }
 
+    private func respondError(_ error: InternalError, topic: String, requestId: RPCID) {
```
llbartekll, line 48: would it make sense if this method is async?

Sources/Auth/Types/Error/ExternalError.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+public enum ExternalError: Codable, Equatable, Error {
```
llbartekll, line 3: maybe PeerError?

Example/IntegrationTests/Auth/AuthTests.swift
llbartekll, line 82: maybe 5, 10 seconds?

Sources/Auth/AuthClient.swift
```diff
@@ -15,8 +15,8 @@ public class AuthClient {
         authRequestPublisherSubject.eraseToAnyPublisher()
     }
 
-    private var authResponsePublisherSubject = PassthroughSubject<(id: RPCID, result: Result<Cacao, ErrorCode>), Never>()
-    public var authResponsePublisher: AnyPublisher<(id: RPCID, result: Result<Cacao, ErrorCode>), Never> {
+    private var authResponsePublisherSubject = PassthroughSubject<(id: RPCID, result: Result<Cacao, InternalError>), Never>()
+    public var authResponsePublisher: AnyPublisher<(id: RPCID, result: Result<Cacao, InternalError>), Never> {
```
llbartekll, line 19: I think the error can be a protocol error.
consider case when wallet rejects a request - user initiated

llbartekll: so if it can be InternalError and ExternalError maybe we could have only one public error like: AuthError?

llbartekll (commented):
I am not convinced about errors separation but for now have no suggestions, maybe let's discuss it?

Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
flypaper0: there is no async context. I think it will be the same

Sources/Auth/Types/Error/ExternalError.swift
flypaper0: My suggestion: ExternalError - errors which comes from outside. InternalError - errors which could occur during SDK execution

llbartekll (approved):

PR #452, opened 2022-08-19
[Auth] #415 auth instance

### LINKED ISSUE #415: Auth Instance wrapper
follow the same pattern as in Sign

llbartekll:
# Description
Add Auth singletone instance that wraps auth client

Resolves #415 

## How Has This Been Tested?
- needs to be tested in example app [#451](https://github.com/WalletConnect/WalletConnectSwiftV2/issues/451)

## Due Dilligence

* [ ] Breaking change


Sources/Auth/Auth.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import WalletConnectRelay
+import Combine
+
+public class Auth {
+
+    public static var instance: AuthClient = {
+        guard let metadata = Auth.metadata else {
+            fatalError("Error - you must call Sign.configure(_:) before accessing the shared instance.")
```
flypaper0, line 9: Auth.configure(_:)

Sources/Auth/Auth.swift
```diff
@@ -0,0 +1,26 @@
...
+            relayClient: Relay.instance)
+    }()
+
+    private static var metadata: AppMetadata?
+    private static var account: Account?
+
+    private init() { }
+
+    static public func configure(metadata: AppMetadata, account: Account?) {
+        Auth.metadata = metadata
```
flypaper0, line 23: mb config if more than 1 property?

Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
```diff
@@ -19,15 +19,15 @@ class WalletRequestSubscriber {
...
+            subscribeForRequest()
+        }
     }
 
     private func subscribeForRequest() {
-        guard let address = address else {
-            logger.warn("unexpected request")
-            return
-        }
+        guard let address = address else {return}
```
flypaper0, line 28: 👍

Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
```diff
@@ -19,15 +19,15 @@ class WalletRequestSubscriber {
         self.logger = logger
         self.address = address
         self.messageFormatter = messageFormatter
-        subscribeForRequest()
+        if address != nil {
```
flypaper0, line 22: we checking this in subscribeForRequest

Sources/Auth/Auth.swift
llbartekll: relay have more and we do not use config.
you removed the Config and actually I liked it

flypaper0 (approved):

Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
llbartekll: true it is redundant, will remove it

PR #461, opened 2022-08-23
[Auth] Test respond error

### LINKED ISSUE #459: [Auth] handle error response

llbartekll:
# Description
- handle peer error response 
- add testUserRespondError
- add reject method to public interface

Resolves #459 #460

## How Has This Been Tested?
- integration tests

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -4,6 +4,7 @@ import WalletConnectUtils
 import WalletConnectPairing
 import WalletConnectRelay
 
+
 public class AuthClient {
     enum Errors: Error {
         case malformedPairingURI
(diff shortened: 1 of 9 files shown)


Sources/Auth/AuthClient.swift
```diff
@@ -87,9 +88,13 @@ public class AuthClient {
...
     }
 
-    public func respond(requestId: RPCID, result: Result<CacaoSignature, Never>) async throws {
+    public func respond(requestId: RPCID, signature: CacaoSignature) async throws {
         guard let account = account else { throw Errors.unknownWalletAddress }
-        try await walletRespondService.respond(requestId: requestId, result: result, account: account)
+        try await walletRespondService.respond(requestId: requestId, signature: signature, account: account)
+    }
+
+    public func reject(requestId: RPCID) async throws {
```
flypaper0, line 96: Good job!

Sources/Auth/Services/App/AppRespondSubscriber.swift
```diff
@@ -28,16 +28,23 @@ class AppRespondSubscriber {
...
             guard
-                let requestId = subscriptionPayload.response.id,
+                let requestId = response.id,
                 let request = rpcHistory.get(recordId: requestId)?.request,
                 let requestParams = request.params, request.method == "wc_authRequest"
             else { return }
 
             networkingInteractor.unsubscribe(topic: subscriptionPayload.topic)
 
+            if let errorResponse = response.error,
```
flypaper0, line 40: guard?

flypaper0 (approved):

Sources/Auth/Services/App/AppRespondSubscriber.swift
llbartekll: if we have guard and response is not error the function would return here but we want to handle error and success response right?

llbartekll: > We sending error code in reject method, and recover AuthError with code here. Right?

correct

PR #475, opened 2022-08-30
[Interface] Replace string with WalletConnectURI

### LINKED ISSUE #465: [API] Replace string URI with WalletConnectURI object

flypaper0:
# Description
- String replaced with WalletConnectURI
- Lint

Resolves #465

## How Has This Been Tested?
Run UI tests

## Due Dilligence

* [x] Breaking change

### DIFF
--- a/Example/ExampleApp/SceneDelegate.swift
+++ b/Example/ExampleApp/SceneDelegate.swift
@@ -46,7 +46,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         vc.onClientConnected = {
             Task {
                 do {
-                    try await Sign.instance.pair(uri: wcUri)
+                    try await Sign.instance.pair(uri: WalletConnectURI(string: wcUri)!)
                 } catch {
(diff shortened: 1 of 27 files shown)


Example/ExampleApp/SceneDelegate.swift
```diff
@@ -46,7 +46,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         vc.onClientConnected = {
             Task {
                 do {
-                    try await Sign.instance.pair(uri: wcUri)
+                    try await Sign.instance.pair(uri: WalletConnectURI(string: wcUri)!)
```
llbartekll, line 49: should we add the check: if uri is for Sign API or something like that?

llbartekll (approved):
I am checking:
- Breaking change
- Requires a documentation update


as I believe both are required.

flypaper0:
Update Docs issue
https://github.com/WalletConnect/WalletConnectSwiftV2/issues/479

Issue #447, opened 2022-08-18
[Sign] Allow to remove pairing


llbartekll:
Update disconnect method to follow the specs:

```
  // for either to disconnect a session or pairing
  public abstract disconnect(params: {
    topic: string;
    reason: Reason;
  }): Promise<void>;

```

PR #486, opened 2022-09-05
[Auth, Sign] #485 remove pairing

### LINKED ISSUE #485: [Auth] Allow to remove pairing

llbartekll:
# Description
- Add disconnect() method to auth client api. Removes pairing
- Add disconnect service for Sign

Resolves #485 #447 
## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectSign/Engine/Common/DeletePairingService.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+import WalletConnectKMS
+import WalletConnectUtils
+import WalletConnectPairing
+
+class DeletePairingService {
```
flypaper0, line 6: Can we move it in WalletConnectPairing package?

flypaper0: oh, I see. Whey using different NetworkInteractors..

llbartekll: yeah exactly, even the protocol method is the same

Sources/WalletConnectSign/Engine/Common/DisconnectService.swift
```diff
@@ -0,0 +1,33 @@
...
+        self.deleteSessionService = deleteSessionService
+        self.pairingStorage = pairingStorage
+        self.sessionStorage = sessionStorage
+    }
+
+    func disconnect(topic: String) async throws {
+        if pairingStorage.hasPairing(forTopic: topic) {
+            try await deletePairingService.delete(topic: topic)
+        } else if sessionStorage.hasSession(forTopic: topic) {
+            try await deleteSessionService.delete(topic: topic)
```
flypaper0, line 28: Pairing and a session at the same time? I mean: do we need to clear pairing and session together on disconnect call?

flypaper0: if so, I think is better to have 1 DeleteSessionService, which clean pairing and session storage

llbartekll: one topic corresponds to one pairing or one session and we have a common public method for deletion.
So we either delete a pairing or a session.

flypaper0 (approved):

flypaper0 (approved):

PR #492, opened 2022-09-07
[CI] Host URL arg

flypaper0:
# Description
RELAY_URL env variable could be passed via GH actions script

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


.github/actions/ci/action.yml
```diff
@@ -27,7 +27,9 @@ runs:
         -project Example/ExampleApp.xcodeproj \ 
         -scheme IntegrationTests \
         -clonedSourcePackagesDirPath SourcePackagesCache \
-        -destination 'platform=iOS Simulator,name=iPhone 13' test"
+        -destination 'platform=iOS Simulator,name=iPhone 13' \ 
+        RELAY_HOST='relay.walletconnect.com' \
+        test"
```
arein, line 32: nit (we can do this in a follow up PR) but it would be nice to make another input at the top of the file with a default value of `relay.walletconnect.com` to allow callers to override

arein (approved):

.github/actions/ci/action.yml
flypaper0: relay.walletconnect.com is the default value (if not passing RELAY_HOST param). Is it enough?

llbartekll (approved):

PR #506, opened 2022-09-14
[Relay] Retry on connection error

flypaper0:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectRelay/RelayClient.swift
+++ b/Sources/WalletConnectRelay/RelayClient.swift
@@ -30,11 +30,10 @@ public final class RelayClient {
     }
 
     public var socketConnectionStatusPublisher: AnyPublisher<SocketConnectionStatus, Never> {
-        socketConnectionStatusPublisherSubject.eraseToAnyPublisher()
+        dispatcher.socketConnectionStatusPublisher
     }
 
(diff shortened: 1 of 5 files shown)


Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -291,4 +294,34 @@ public final class RelayClient {
...
+                if let error = error {
+                    continuation.resume(throwing: error)
+                } else {
+                    continuation.resume(returning: ())
+                }
+            }
+        }
+    }
+
+    private func protectedSend(_ string: String, timeout: Int = 5, completion: @escaping (Error?) -> Void) {
```
llbartekll, line 310: I would move it to dispatcher.
Relay client is responsible for client<->relay interactions - subscribe/publish etc.
dispatcher is responsible for managing behaviour/policy of sending requests - unaware of the protocol methods etc.

and `protectedSend` to me sounds more like dispatcher responsibility.

llbartekll (changes requested):
looks good, I would just move the protectedSend method to dispatcher

llbartekll (approved):

PR #557, opened 2022-10-26
[Chat] PubKey removed from invite

flypaper0:
# Description
Invite method updated according specs 

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


flypaper0:
> I think pubkey in invite makes a lot of sense in outbound peer discovery(qr code). Maybe we should update specs instead?

I agree, but account based invite is also useful. We should support both methods ig

llbartekll:
> > I think pubkey in invite makes a lot of sense in outbound peer discovery(qr code). Maybe we should update specs instead?
> 
> I agree, but account based invite is also useful. We should support both methods ig

Do you want to expose a second metod then?

PR #577, opened 2022-11-09
[Sign] proposal public for uniformity

kirill09:
# Description

in the kotlin version of the sdk proposal is public. For consistency, the swift proposal is changed to public

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


llbartekll:
hey @kirill09  Any particular reason you need this model public?
take look at `Session.Proposal` object it's a public version of the session proposal

kirill09:
Hi @llbartekll Session.Proposal is public but the proposal parameter is hidden.  The public proposal is needed to get theProposalKey, protocol, data,

llbartekll:
> Hi @llbartekll Session.Proposal is public but the proposal parameter is hidden. The public proposal is needed to get theProposalKey, protocol, data,

but you don't need protocol and data and for convenience you can reference Proposal by id

kirill09:
> 

I analyzed again what we use. We need publicKey for rejectSession, approveSession.

kirill09:
> > Hi @llbartekll Session.Proposal is public but the proposal parameter is hidden. The public proposal is needed to get theProposalKey, protocol, data,
> 
> but you don't need protocol and data and for convenience you can reference Proposal by id

I analyzed again what we use. We need publicKey for rejectSession, approveSession.

llbartekll:
> I analyzed again what we use. We need publicKey for rejectSession, approveSession.

use id from Session.Proposal when calling
```approve(proposalId: String,...)```

you can take a look at our example wallet, all works there

kirill09:
> 



> > I analyzed again what we use. We need publicKey for rejectSession, approveSession.
> 
> use id from Session.Proposal when calling `approve(proposalId: String,...)`
> 
> you can take a look at our example wallet, all works there

Hi. This is a great solution but for android there is no Session.Proposal it is not very convenient to use different implementations with a single code base like react native or flutter

llbartekll:
> Hi. This is a great solution but for android there is no Session.Proposal it is not very convenient to use different implementations with a single code base like react native or flutter

but still making this property public does not make a lot of difference as non of the client's public methods use it.
do you think that for your needs you can couple somehow `Session.Proposal` with android's session proposal?

PR #578, opened 2022-11-09
[Sign] For the error we use response error

kirill09:
# Description

Incorrect response on rejection of the request

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


jackromo888:
Thanks

Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
```diff
@@ -59,11 +59,16 @@ final class SessionEngine {
...
         guard sessionStore.hasSession(forTopic: topic) else {
             throw Errors.sessionNotFound(topic: topic)
         }
-        let response = RPCResponse(id: requestId, result: response)
+      
+      var response = RPCResponse(id: requestId, result: result)
+      
+      if case .error(let value) = result {
+        response = RPCResponse(id: requestId, error: value);
+      }
```
llbartekll, line 71: hey @kirill09 that's a good point, the initialiser is incorrect for an error response but after your changes tests are not passing. 

Did you encounter any problems with current implementation?

kirill09: hi @llbartekll In the current implementation, I was getting an RPCResponse result object and it had an error in it. 
Which was causing the wrong behavior. 
Actions didn't get to the point of running tests. I think these errors are not related to my changes. Can you restart the actions?

llbartekll: it's interesting because the response actually decodes to error. take look at this test, it passes:
https://github.com/WalletConnect/WalletConnectSwiftV2/blob/84197cd011da4b1986fc7a431cdf3f0f32199db1/Example/IntegrationTests/Sign/SignClientTests.swift#L247
and after your changes it does not, i will take a closer look at it

kirill09: @llbartekll Perhaps the latest commit will fix the tests

llbartekll:
>  I was getting an RPCResponse result object and it had an error in it.

was it unexpected? RPCResponse can contain an error object.

can you take a look again at this test https://github.com/WalletConnect/WalletConnectSwiftV2/blob/84197cd011da4b1986fc7a431cdf3f0f32199db1/Example/IntegrationTests/Sign/SignClientTests.swift#L247
when a wallet responds with an error sessionResponsePublisher on dapp publishes an error

llbartekll:
@kirill09 take a look at this pr:
https://github.com/WalletConnect/WalletConnectSwiftV2/pull/600
I think it will solve your issue

llbartekll:
if not objections, I will close this pr

kirill09:
> @kirill09 take a look at this pr: #600 I think it will solve your issue

That sounds like the right thing to do. I'll try to check it out in a couple of days.

PR #626, opened 2022-12-06
Extract wallet app for chat auth + architecture adjustments

alexander-lsvk:
Few words about architecture adjustments I wanted to introduce.

In the current implementation I was embarrassed that the roles of the **interactor** and **presenter** layers are a bit mixed.  Both of them contain some logic handling, but mostly **interactor** itself right now looks a bit redundant, because its code can be easily moved to the presenter and basically it won't affect anything. It's not only about cleanliness but also about entry level to the example apps so people (like me) can understand easily how it works. 

As for me the role of **presenter** (if we are talking about `ObservableObject`) in the configuration where we have **interactor**, should not contain any logic: like calling **router**, subscribing on some logic related topics, etc. - **presenter** should be just a UI state snapshot.

But at the same time SwiftUI pushes us to use 2 ways binding between **presenter** and **view**, what makes us call **presenter** directly from the **view**, which can lead to the state/data inconsistency. The problem is such connection is difficult to test and it breaks the clean responsibilities of each layer.

So solution might be to drop a 2 ways binding between **presenter** and **view**, and keep only 1 direction for the data flow.

In this case 
- **Interactor** is the only layer responsible for the all logic related stuff, like calling SDK methods, internal logic, calling router and triggering presenter to update UI - all logic in the module can be easily mocked.
- **Presenter** basically knows nothing about **interactor** and **view** (except SwiftUI bindings) and responsible for the data formatting
- **View** gets UI ready data from **presenter** and sends all logic related user interactions to the **interactor**

This relations can be easily illustrated using the scheme below (the implementation is in `WalletModule`).

Untitled

Otherwise I'd think about dropping **interactor** layer, and keep everything in **presenter**. Since this is an example app probably it doesn't require to be so precise in the layer responsibilities, cuz the main goal is to be simple for people to understand how it works.

Anyway let's discuss it below.

## Due Dilligence

* [ ] Breaking change


Example/ChatWallet/ApplicationLayer/AppDelegate.swift
```diff
@@ -0,0 +1,19 @@
+import UIKit
+
+@main
+final class AppDelegate: UIResponder, UIApplicationDelegate {
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        // Override point for customization after application launch.
+        print("lala")
```
llbartekll, line 7: can we remove it? xd

Example/ChatWallet/ApplicationLayer/Configurator/AppearanceConfigurator.swift
```diff
@@ -0,0 +1,19 @@
+import UIKit
```
llbartekll, line 1: all the files are in Example/chatWallet dir. Chat app is not a wallet. I think Example/Chat would be good

Example/ChatWallet/ApplicationLayer/SceneDelegate.swift
```diff
@@ -0,0 +1,43 @@
...
+          return sceneConfig
+      }
+
+    func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
+        guard let windowScene = (scene as? UIWindowScene) else { return }
+
+        window = UIWindow(windowScene: windowScene)
+        window?.makeKeyAndVisible()
+
+        print("Lolo")
```
llbartekll, line 31: same

llbartekll (commented):
approving but let's wait for Artur's approval as well, he have more insight on chat app.

Example/ChatWallet/Common/Helpers/UIPasteboardWrapper.swift
```diff
@@ -0,0 +1,15 @@
+//
```
flypaper0, line 1: normally we removing all file headers, but not critical

Example/ChatWallet/PresentationLayer/Wallet/Wallet/WalletView.swift
```diff
@@ -0,0 +1,65 @@
+import SwiftUI
+
+struct WalletView: View {
+    let interactor: WalletInteractorProtocol
```
flypaper0, line 4: In classic VIPER patter Interactor is to interact with services.  
I agree, now interactor is looks like redundant but this is only because the app on the initial stage and we does not have a lot services. In future we will have more storage logic, push notification services, network services etc.

If we injecting interactor in the View layer, that means what we are mixing service logic with view specific logic (filtering, updating view, reloading etc). 

If you fill what interactor is redundant I think is better to just remove it (if it not needed for this module), then transfer view events into it instead of presenter.

Example/ChatWallet/ApplicationLayer/AppDelegate.swift
alexander-lsvk: Yeah, removed already, you shouldn't look there haha 😂

Example/ChatWallet/Common/Extensions/UIKit/UIColor.swift
```diff
@@ -0,0 +1,20 @@
+import UIKit
```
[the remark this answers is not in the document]
alexander-lsvk: Just to make wallet work and share the thoughts about architecture, so basically things to review are under `ChatWallet/PresentationLayer/Wallet`

Example/ChatWallet/PresentationLayer/Wallet/Wallet/WalletView.swift
alexander-lsvk: If you tap, f.i. on some filter button, you pass you selection from **Presenter** as a parameter to the **Interactor**. **Interactor** does its job and updates view via **Presenter**. Is there some mixing logic with **View**?

My point was mostly about the fact that app's business logic is spread between **Presenter** and **Interactor** and it doesn't depend on how much services, etc. Since **Presenter** in **SwiftUI** is a bit different, I wouldn't try to fit some "classic VIPER" and stuff, just stick to some clean solution.

But yeah removing the **Interactor** I believe is only a temporary solution. As you've mentioned if we get a bigger app, **Presenter** will grow up and things get messy.

PR #638, opened 2022-12-17
Web3Wallet implementation + integration tests

alexander-lsvk:
# Description

Web3Wallet proxy implementation according to the interface described [here](https://gist.github.com/pedrouid/4067a262550ffb80ea93ab21cedba2fb) 

* [x] Breaking change


Sources/Web3Wallet/Web3Wallet.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    private init() { }
+
+    /// Wallet instance wallet config method.
+    /// - Parameters:
+    ///   - metadata: App metadata
+    ///   - signerFactory: Auth signers factory
+    static public func configure(metadata: AppMetadata, signerFactory: SignerFactory) {
+        Pair.configure(metadata: metadata)
+        Sign.configure(metadata: metadata)
```
flypaper0, line 46: Sign.configure is deprecated, no need to call it

Sources/Web3Wallet/Web3Wallet.swift
```diff
@@ -0,0 +1,51 @@
...
+            fatalError("Error - you must call Wallet.configure(_:) before accessing the shared instance.")
+        }
+        
+        return Web3WalletClientFactory.create(
+            metadata: Pair.metadata,
+            projectId: Networking.projectId,
+            signerFactory: config.signerFactory,
+            networkingClient: Networking.interactor,
+            pairingRegisterer: Pair.registerer,
+            pairingClient: Pair.instance as! PairingClient
```
flypaper0, line 32: @llbartekll as I remember we found the way how to avoid force casts for pairing client. Or we decided to leave it like that?

Example/IntegrationTests/Web3Wallet/Web3WalletClientTest.swift
```diff
@@ -0,0 +1,402 @@
...
+import WalletConnectUtils
+import JSONRPC
+@testable import WalletConnectKMS
+@testable import WalletConnectSign
+@testable import Web3Wallet
+@testable import WalletConnectRelay
+import WalletConnectPairing
+import WalletConnectNetworking
+
+final class Web3WalletClientTests: XCTestCase {
```
flypaper0, line 11: is it copy or Sign and Auth integration tests? Do we need this?

Sources/Web3Wallet/Web3Wallet.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import Combine
+
+import Auth
+import WalletConnectSign
```
flypaper0, line 5: please check how we are doing @_exported imports in other packages. Otherwise it will break cocoapods releases

Sources/Web3Wallet/Web3Wallet.swift
llbartekll: We have fixed it for Networking having two vars: `instance` and `interactor` I think we could do the same with pairing

Sources/Web3Wallet/Web3Wallet.swift
flypaper0: just checked, this also need to do for Push SDK @llbartekll

Sources/Web3Wallet/Web3Wallet.swift
```diff
@@ -0,0 +1,51 @@
...
+/// Web3Wallet.instance.getSessions()
+/// ```
+public class Web3Wallet {
+    /// Web3Wallett client instance
+    public static var instance: Web3WalletClient = {
+        guard let config = Web3Wallet.config else {
+            fatalError("Error - you must call Wallet.configure(_:) before accessing the shared instance.")
+        }
+        
+        return Web3WalletClientFactory.create(
```
flypaper0, line 26: Is it possible to create instance with that initializer? 
```swift
Web3WalletClientFactory.create(sign: Sign.instance, auth: Auth.instance)
```

flypaper0: Because we are already configured Sign and Auth

Example/IntegrationTests/Web3Wallet/Web3WalletClientTest.swift
llbartekll: to me it looks like a copy of just a sign tests.
running sign and web3wallet tests together will significantly increase testing on CI without many benefits.

what do you think to just unit test web3walletClient to ensure it calls sign and auth methods?

combining sign and auth tests into one class may be a bit monolithic approach 🤔

llbartekll: it would require sign and auth clients mocks

Sources/Web3Wallet/Web3WalletImports.swift
```diff
@@ -0,0 +1,3 @@
+#if !CocoaPods
+@_exported import WalletConnectPairing
```
flypaper0, line 2: also need to remove all packages imports from all files

flypaper0: Cocoapods putting all sources in single framework, so it doesn't know what is Auth, WalletConnectSign etc

Need to remove all 

```swift
import WalletConnectSign
import Auth
```
And imports should be
```swift
@_exported import Auth
@_exported import WalletConnectSign
```

Agree it looks like a little wired but supporting both SPM and Cocoapods is a kind of pain

alexander-lsvk:
@llbartekll @flypaper0 I added unit tests for `Web3Wallet` in the last commit. It required adding abstraction levels for `Auth`, `Sign` and `Pair` -> `AuthClientProtocol`, `SignClientProtocol`, `PairClientProtocol`.

I'm not really satisfied with this `*Protocol` suffix, I'd be more pleased to have protocol named, i.e. `SignClient`, and then `SignClientMock` and `SignClient*Somethings*` for implementation. But the question what this *Something* should be 🤔

Any suggestions?

llbartekll (approved):
Looks good to me.

flypaper0:
> @llbartekll @flypaper0 I added unit tests for `Web3Wallet` in the last commit. It required adding abstraction levels for `Auth`, `Sign` and `Pair` -> `AuthClientProtocol`, `SignClientProtocol`, `PairClientProtocol`.
> 
> I'm not really satisfied with this `*Protocol` suffix, I'd be more pleased to have protocol named, i.e. `SignClient`, and then `SignClientMock` and `SignClient*Something*` for implementation. But the question is what this _Something_ should be 🤔
> 
> Any suggestions?

For me all namings is good, but it must be consistant across project and all our sdk's

flypaper0 (approved):
lgtm!

PR #668, opened 2023-01-16
Extend 'Session' with 'pairingTopic'

alexander-lsvk:
# Description

Base on https://walletconnect.slack.com/archives/C03RR5C0F7X/p1673428495333859

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectSign/Services/App/AppProposeService.swift
+++ b/Sources/WalletConnectSign/Services/App/AppProposeService.swift
@@ -26,12 +26,12 @@ final class AppProposeService {
         let proposer = Participant(
             publicKey: publicKey.hexRepresentation,
             metadata: metadata)
+        
         let proposal = SessionProposal(
             relays: [relay],
             proposer: proposer,
-            pairingTopic: pairingTopic,
             requiredNamespaces: namespaces)
-
+        
         let request = RPCRequest(method: protocolMethod.method, params: proposal)
         try await networkingInteractor.requestNetworkAck(request, topic: pairingTopic, protocolMethod: protocolMethod)
     }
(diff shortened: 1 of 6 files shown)


Sources/WalletConnectSign/Services/App/AppProposeService.swift
```diff
@@ -29,6 +29,7 @@ final class AppProposeService {
         let proposal = SessionProposal(
             relays: [relay],
             proposer: proposer,
+            pairingTopic: pairingTopic,
```
llbartekll, line 32: SessionProposal does not need this param, we are not relying on protocol object but on pairing topic the proposal has been received on

Sources/WalletConnectSign/Session.swift
```diff
@@ -17,6 +17,7 @@ extension Session {
 
     public struct Proposal: Equatable {
         public var id: String
+        public let pairingTopic: String
```
llbartekll, line 20: this is correct

llbartekll (changes requested):
hey you need to take topic from RequestSubscriptionPayload in approve engine 
```swift
    func setupRequestSubscriptions() {
        pairingRegisterer.register(method: SessionProposeProtocolMethod())
            .sink { [unowned self] (payload: RequestSubscriptionPayload<SessionType.ProposeParams>) in
                handleSessionProposeRequest(payload: payload)
            }.store(in: &publishers)
```

Sources/WalletConnectSign/Services/App/AppProposeService.swift
llbartekll: it's not secure to rely on what proposer is claiming here

llbartekll:
@alexander-lsvk if you have questions, ping me for a call

Tests/WalletConnectSignTests/Stub/Stubs.swift
```diff
@@ -77,12 +77,11 @@ extension RPCRequest {
 }
 
 extension SessionProposal {
-    static func stub(proposerPubKey: String = "") -> SessionProposal {
+    static func stub(proposerPubKey: String = "", topic: String = "") -> SessionProposal {
```
llbartekll, line 80: looks like it's not used

PR #683, opened 2023-01-23
[Echo] allow to configure echo host

llbartekll:
Allows to configure echo server host for projects that want self hosted echo server

### DIFF
--- a/Sources/WalletConnectEcho/Echo.swift
+++ b/Sources/WalletConnectEcho/Echo.swift
@@ -10,7 +10,8 @@ public class Echo {
 
         return EchoClientFactory.create(
             projectId: Networking.projectId,
-            clientId: config.clientId)
+            clientId: config.clientId,
+            echoHost: config.echoHost)
     }()
 
     private static var config: Config?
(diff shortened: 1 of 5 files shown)


Sources/WalletConnectEcho/Echo.swift
```diff
@@ -19,7 +20,10 @@ public class Echo {
 
     /// Echo instance config method
     /// - Parameter clientId: https://github.com/WalletConnect/walletconnect-docs/blob/main/docs/specs/clients/core/relay/relay-client-auth.md#overview
-    static public func configure(clientId: String) {
-        Echo.config = Echo.Config(clientId: clientId)
+    static public func configure(
+        clientId: String,
+        echoHost: String = "echo.walletconnect.com"
```
flypaper0, line 25: may be constant in Echo SDK? to reuse it

flypaper0 (approved):

Sources/WalletConnectEcho/Echo.swift
llbartekll: good point, I will fix it

alexander-lsvk (approved):

PR #688, opened 2023-01-25
[Sign] CAIP-25

alexander-lsvk:
## Due Dilligence

* [x] Breaking change


flypaper0:
Do we also need to change namespaces validation code?
Will be nice to add some tests for optional namespaces in NamespaceValidationTests.swift

alexander-lsvk:
> Do we also need to change namespaces validation code?
> Will be nice to add some tests for optional namespaces in NamespaceValidationTests.swift

Yeah, I mentioned it in Slack, if implementation looks good to you, I'll push some tests 👌🏻

Sources/WalletConnectSign/Namespace.swift
```diff
@@ -47,6 +57,17 @@ enum Namespace {
...
         }
     }
+    
+    static func validate(_ namespaces: [String: OptionalNamespace]) throws {
+        for (_, namespace) in namespaces {
+            if namespace.methods.isEmpty {
+                throw WalletConnectError.unsupportedNamespace(.unsupportedMethods)
+            }
+            if namespace.events.isEmpty {
+                throw WalletConnectError.unsupportedNamespace(.unsupportedEvents)
```
llbartekll, line 67: is it always required to have an event in optional namespace?

alexander-lsvk: Screenshot 2023-01-30 at 11 38 28
In the doc the whole namespace structure is covered regardless of being `required` or `optional` so I think so.

llbartekll: should the optional namespace contain the chains too then?
I don't think events and methods cannot be empty but let's discuss it today on the w3w meeting.

flypaper0 (approved):

flypaper0:
will be good to add a few negative validation tests

llbartekll (approved):
should we merge it after js/swift/kotlin tests?
there may be some issues discovered

llbartekll (approved):

PR #691, opened 2023-01-27
Ensure application state read from main thread

### LINKED ISSUE #685: AppStateObserver calling `applicationState` in background thread
**Describe the bug**
AppStateObserver calling code not on main thread. (See screenshot)

**SDK Version**
1.3.0

**To Reproduce**
Go to background to disconnect socket, or wait until socket disconnects from dapp

**Expected behavior**
The warning should not appear

**Screenshots**

Screenshot 2023-01-23 at 16 45 35

**Device (please complete the following information):**
 - Device: Simulator
 - OS: iOS16.2

basememara:
# Description

When providing a background thread to the socket factory instance, the underlying `AppStateObserver.currentState` logic calls `UIApplication.shared.applicationState` which needs to be called on the main thread.

The primary change is within the `AppStateObserving` protocol: the `currentState` property was made `async` so implementors can add the `@MainActor` property wrapper or manually run a `Task` on the main thread.

Resolves #685

## How Has This Been Tested?

1. For `DefaultSocketFactory` in the `ExampleApp`, a background thread was provided to the web socket instance:

```swift
let socket = WebSocket(url: url)
socket.callbackQueue = DispatchQueue.global()
return socket
```

The runtime warning in Xcode was observed stating `UIApplication.applicationState must be used from the main thread only`. The warning went away after the fix including going to and from the background state.

2. Unit tests had succeeded, but uncovered an issue in `RelayerTests: AutomaticSocketConnectionHandlerTests testReconnectOnDisconnectForeground` that required a change.

Because the tests assume that the socket is synchronous, the reconnect is assumed to happen instantly after disconnect. I had to simulate some time for the reconnect before checking the `isConnect` flag within the test:

```swift
Task { @MainActor in
    // Simulate time for reconnect
    XCTAssertTrue(webSocketSession.isConnected)
}
```

A better (but larger) fix for tests may be to make `WebSocketMock` asynchronous to more closely simulate real world connections. The tests would be more robust but would require a lot of changes such as making `WebSocketConnecting.isConnect` an `async` property or maybe tests should check for connection state within `socketConnectionStatusPublisher` instead.

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Tests/RelayerTests/AutomaticSocketConnectionHandlerTests.swift
+++ b/Tests/RelayerTests/AutomaticSocketConnectionHandlerTests.swift
@@ -55,12 +55,15 @@ final class AutomaticSocketConnectionHandlerTests: XCTestCase {
         XCTAssertFalse(webSocketSession.isConnected)
     }
 
-    func testReconnectOnDisconnectForeground() {
+    func testReconnectOnDisconnectForeground() async {
         appStateObserver.currentState = .foreground
         XCTAssertTrue(webSocketSession.isConnected)
         webSocketSession.disconnect()
         sut.handleDisconnection()
-        XCTAssertTrue(webSocketSession.isConnected)
+        Task { @MainActor in
+            // Simulate time for reconnect
+            XCTAssertTrue(webSocketSession.isConnected)
+        }
     }
 
     func testReconnectOnDisconnectBackground() {
--- a/Tests/RelayerTests/AutomaticSocketConnectionHandlerTests.swift
+++ b/Tests/RelayerTests/AutomaticSocketConnectionHandlerTests.swift
@@ -55,15 +55,18 @@ final class AutomaticSocketConnectionHandlerTests: XCTestCase {
         XCTAssertFalse(webSocketSession.isConnected)
     }
 
-    func testReconnectOnDisconnectForeground() async {
+    func testReconnectOnDisconnectForeground() {
+        let expectation = expectation(description: "expects to reconnect on disconnect")
         appStateObserver.currentState = .foreground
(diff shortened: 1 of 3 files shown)


basememara:
The CI got stuck on the integration test, although it passes locally when I add my own project ID to the `.xcconfig`. Is there an issue with the project for the CI or is there something extra I needed to do?

Tests/RelayerTests/AutomaticSocketConnectionHandlerTests.swift
```diff
@@ -55,12 +55,15 @@ final class AutomaticSocketConnectionHandlerTests: XCTestCase {
...
+    func testReconnectOnDisconnectForeground() async {
         appStateObserver.currentState = .foreground
         XCTAssertTrue(webSocketSession.isConnected)
         webSocketSession.disconnect()
         sut.handleDisconnection()
-        XCTAssertTrue(webSocketSession.isConnected)
+        Task { @MainActor in
+            // Simulate time for reconnect
+            XCTAssertTrue(webSocketSession.isConnected)
+        }
```
llbartekll, line 66: I think this test will always pass even if you comment out `sut.handleDisconnection()`
what about adding an expectation?

basememara: Good catch! I was thinking `await Task.sleep` but what I have is a separate task not awaited for. I've added an `expectation` like you suggested and asserting / fulfilling within the async execution.

llbartekll (approved):

flypaper0:
closed with https://github.com/WalletConnect/WalletConnectSwiftV2/pull/746

PR #693, opened 2023-01-27
[CI] Fastlane integration

flypaper0:
# Description
- CI using Fastlane scripts from makefile
- TestFlight release scripts

# For TestFlight release:
1. Install fastlane: 
`make install_env`

2. Setup environment
```
echo 'export FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD={password}’ >> ~/.zshenv
echo 'export MATCH_PASSWORD={password}' >> ~/.zshenv
```
Match password I'll share. App specific password can be generated on apple.id website. 

3. Release
```
make release_showcase APPLE_ID={your apple id}
make release_walletapp APPLE_ID={your apple id}
```

## Due Dilligence

* [ ] Breaking change


Makefile
```diff
@@ -2,7 +2,34 @@ XCODE_USER_TEMPLATES_DIR=/Applications/Xcode.app/Contents/Developer/Library/Xcod
...
+	sudo gem install fastlane --no-document
+endif		
+	@echo "All dependencies was installed"
+
+build_dapp:
+	fastlane build scheme:DApp
+
+build_wallet:
+	fastlane build scheme:WalletApp
+
```
alexander-lsvk, line 24: If I understand correctly it will build apps with the every CI check?
If so I think it's a little bit overhead for the resources and time. This check would make more sense to me when merging to `master` i.e.

flypaper0: Our tests is running in parallel, so CI time depends on longes test case. Now it is Integration tests. I can switch off this case if it make sense for you

fastlane/Fastfile
```diff
@@ -13,29 +13,70 @@
...
+      type: "appstore",
+      app_identifier: ENV["APP_IDENTIFIER"],
+      git_url: "https://github.com/WalletConnect/match-swift.git",
+      username: options[:username],
+    )
+    number = latest_testflight_build_number(
+      app_identifier: ENV["APP_IDENTIFIER"],
+      username: options[:username],
+    )
+    increment_build_number(
```
alexander-lsvk, line 65: 😍

llbartekll (approved):
sweet!

alexander-lsvk (approved):

llbartekll:
@flypaper0 I think we should test it with Infra pipeline before merging.

PR #706, opened 2023-02-06
[Chat] Identity and Invite keys registration

flypaper0:
# Description
- identity key registration
- invite key registration
- resolving identity and invite keys
- integration tests
- WalletConnectJWT package
- Cacao and DID stuff moved to WalletConnectUtills
- WalletConnectSigner package 

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectSigner/Signer/MessageSigner.swift
```diff
@@ -27,22 +31,28 @@ struct MessageSigner: AuthMessageSigner {
     private let signer: EthereumSigner
     private let eip191Verifier: EIP191Verifier
     private let eip1271Verifier: EIP1271Verifier
-    private let messageFormatter: SIWEMessageFormatting
+    private let messageFormatter: SIWECacaoFormatting
```
llbartekll, line 34: it was called SIWEMessageFormatting because siwe message does not know anything about cacao, not sure if it's better name

Sources/WalletConnectSigner/Signer/MessageSigner.swift
```diff
@@ -9,16 +9,20 @@ public protocol MessageSignatureVerifying {
...
+    ) throws -> CacaoSignature
+
+    func sign(payload: CacaoPayload,
         privateKey: Data,
         type: CacaoSignatureType
     ) throws -> CacaoSignature
 }
 
-public typealias AuthMessageSigner = MessageSignatureVerifying & MessageSigning
+public typealias CacaoMessageSigner = MessageSignatureVerifying & MessageSigning
```
llbartekll, line 23: same here CacaoMessageSigner couples this to cacao object but it doesn't have to be wrapped into cacao

Sources/WalletConnectUtils/DID/DIDPKH.swift
```diff
@@ -25,8 +26,15 @@ struct DIDPKH {
...
         self.iss = "\(DIDPKH.didPrefix):\(account.absoluteString)"
         self.account = account
     }
 }
+
+extension Account {
+
+    public var iss: String {
+        return DIDPKH(account: self).iss
+    }
```
llbartekll, line 39: what is this used for? what if you want an issuer in with different did method not pkh?

llbartekll (approved):

Sources/WalletConnectUtils/DID/DIDPKH.swift
flypaper0: I'll rename it into `string` in another PR, to not catch the conflict

Sources/WalletConnectSigner/Signer/MessageSigner.swift
flypaper0: same

PR #721, opened 2023-02-16
[Core] Consume PublishedAt irn_subscription property

llbartekll:
# Description
Consume `publishedAt` irn_subscription property and pass it into RequestSubscriptionPayload that is published by networking interactor

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectRelay/RPC/Methods/Subscription.swift
```diff
@@ -1,12 +1,41 @@
...
+import WalletConnectUtils
 
 struct Subscription: RelayRPC {
 
     struct Params: Codable {
         struct Contents: Codable {
+
             let topic: String
             let message: String
+            let publishedAt: Date
```
flypaper0, line 12: do we need to rewrite encoder code? 

we can parse it as is, make it private and create computed property: 
```swift
private let publishedAt: UInt64
var timestamp: Date {
return Date(milliseconds: publishedAt)
}
```

Sources/WalletConnectRelay/RelayClient.swift
```diff
@@ -226,7 +226,7 @@ public final class RelayClient {
                     try rpcHistory.set(request, forTopic: params.data.topic, emmitedBy: .remote)
                     logger.debug("topic \(params.data.topic)")
                     logger.debug("message: \(params.data.message)")
-                    messagePublisherSubject.send((params.data.topic, params.data.message))
+                    messagePublisherSubject.send((params.data.topic, params.data.message, params.data.publishedAt))
```
flypaper0, line 229: may be rename to timestamp? See comment above

flypaper0 (approved):

PR #725, opened 2023-02-21
Localized error

izakpavel:
# Description

Some types are not available when importing the package. These changes allow:
Creating a Session.Event so it can be sent to the peer.
Displaying a localized error message for WalletConnectError


flypaper0:
lgtm!

PR #740, opened 2023-03-08
Echo register auth

llbartekll:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectEcho/Register/EchoRegisterService.swift
```diff
@@ -1,12 +1,14 @@
...
 actor EchoRegisterService {
     private let httpClient: HTTPClient
     private let projectId: String
     private let clientId: String
     private let logger: ConsoleLogging
     private let environment: APNSEnvironment
+    private let echoAuthenticator: EchoAuthenticating
     // DID method specific identifier
     private var clientIdMutlibase: String {
         return clientId.replacingOccurrences(of: "did:key:", with: "")
```
flypaper0, line 14: may be use DIDKey object?

llbartekll: in what way?
actually clientIdStorage could return DIDKey?

flypaper0: i mean: 
`let clientId: DIDKey`

and after you can get the string with `did(prefix: Bool, variant: DIDKeyVariant)`

I created DIDKey type to avoid of using string keys. Actually good idea to return DIDKey from clientIdStorage

Sources/WalletConnectNetworking/HTTPClient/HTTPService.swift
```diff
@@ -12,6 +12,7 @@ public protocol HTTPService {
     var method: HTTPMethod { get }
     var body: Data? { get }
     var queryParameters: [String: String]? { get }
+    var headerFields: [String: String]? { get }
```
flypaper0, line 15: may be `additionalHeaders`, because it's not represent all request headers as we are setting `Content-Type` separately after

flypaper0 (approved):
lgtm, few comments

Sources/WalletConnectEcho/Register/EchoRegisterService.swift
llbartekll: clientID cannot be DIDkey because we could confuse the variant

Sources/Chat/ProtocolServices/Invitee/InvitationHandlingService.swift
```diff
@@ -122,7 +122,7 @@ private extension InvitationHandlingService {
                     let inviterAccount = try await identityClient.resolveIdentity(iss: claims.iss)
                     // TODO: Should we cache it?
                     let inviteePublicKey = try await identityClient.resolveInvite(account: inviterAccount)
-                    let inviterPublicKey = invite.inviterPublicKey.did(prefix: false, variant: .X25519)
+                    let inviterPublicKey = invite.inviterPublicKey.multibase(variant: .ED25519)
```
flypaper0, line 125: why you changed for `ED25519`? inviterPublic key is creating by `createX25519KeyPair`

llbartekll: was a typo, added did prefix too

PR #741, opened 2023-03-08
[WalletApp] Small fixes

alexander-lsvk:
Throw error when method is not implemented, update UI for requested/optional namespaces

## Due Dilligence

* [ ] Breaking change


Example/WalletApp/PresentationLayer/Wallet/SessionRequest/SessionRequestInteractor.swift
```diff
@@ -3,12 +3,16 @@ import Web3Wallet
...
+        do {
+            let result = try Signer.sign(request: sessionRequest)
+            try await Web3Wallet.instance.respond(
+                topic: sessionRequest.topic,
+                requestId: sessionRequest.id,
+                response: .response(result)
+            )
+        } catch {
+            throw error
+        }
```
llbartekll, line 15: what do we need do catch block here?

llbartekll (approved):

Example/WalletApp/PresentationLayer/Wallet/SessionRequest/SessionRequestInteractor.swift
alexander-lsvk: Just to pass an error on the top level for presenting it with alert on UI

llbartekll: but wouldn't it work the same way if it just throws in line 7 or 8?

PR #744, opened 2023-03-13
[Sign/Web3Wallet] Add pending proposals

alexander-lsvk:
## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectSign/Services/HistoryService.swift
+++ b/Sources/WalletConnectSign/Services/HistoryService.swift
@@ -3,11 +3,21 @@ import Foundation
 final class HistoryService {
 
     private let history: RPCHistory
+    private let proposalPayloadsStore: CodableStore<RequestSubscriptionPayload<SessionType.ProposeParams>>
 
-    init(history: RPCHistory) {
+    init(
+        history: RPCHistory,
+        proposalPayloadsStore: CodableStore<RequestSubscriptionPayload<SessionType.ProposeParams>>
+    ) {
         self.history = history
+        self.proposalPayloadsStore = proposalPayloadsStore
     }
 
+    public func getSessionRequest(id: RPCID) -> Request? {
+        guard let record = history.get(recordId: id) else { return nil }
+        return mapRequestRecord(record)
+    }
+    
     func getPendingRequests() -> [Request] {
         return history.getPending()
             .compactMap { mapRequestRecord($0) }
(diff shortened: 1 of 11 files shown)


Sources/WalletConnectSign/Services/HistoryService.swift
```diff
@@ -39,4 +67,25 @@ private extension HistoryService {
...
+        guard let proposal = try? record.request.params?.get(SessionType.ProposeParams.self)
+        else { return nil }
+        return proposal
+    }
+    
+    func mapProposalRecord(_ record: RPCHistory.Record) -> Session.Proposal? {
+        guard let proposal = try? record.request.params?.get(SessionType.ProposeParams.self)
+        else { return nil }
+        
+        return Session.Proposal(
```
llbartekll, line 81: you can use SessionProposal's `publicRepresentation(pairingTopic: String) -> Session.Proposal` here

Sources/WalletConnectSign/Services/HistoryService.swift
```diff
@@ -17,15 +27,33 @@ final class HistoryService {
...
+                guard let proposalParams = mapProposeParams(record) else {
+                    return nil
+                }
+                return RequestSubscriptionPayload(id: record.id, topic: record.topic, request: proposalParams, publishedAt: Date())
+            }
+        
+        requestSubscriptionPayloads.forEach {
+            let proposal = $0.request
+            proposalPayloadsStore.set($0, forKey: proposal.proposer.publicKey)
+        }
```
llbartekll, line 45: you do it to set values to proposalPayloadsStore on app launch?

Sources/WalletConnectSign/Services/HistoryService.swift
```diff
@@ -17,15 +27,33 @@ final class HistoryService {
...
+    
+    func getPendingProposals() -> [Session.Proposal] {
+        let pendingHistory = history.getPending()
+        
+        let requestSubscriptionPayloads = pendingHistory
+            .compactMap { record -> RequestSubscriptionPayload<SessionType.ProposeParams>? in
+                guard let proposalParams = mapProposeParams(record) else {
+                    return nil
+                }
+                return RequestSubscriptionPayload(id: record.id, topic: record.topic, request: proposalParams, publishedAt: Date())
```
llbartekll, line 39: I don't like we are recreating RequestSubscriptionPayload type here.
could we change type stored by proposalPayloadsStore?

Sources/WalletConnectSign/Services/HistoryService.swift
llbartekll: make sense

Sources/WalletConnectSign/Services/HistoryService.swift
```diff
@@ -17,15 +27,33 @@ final class HistoryService {
...
+            }
+        
+        requestSubscriptionPayloads.forEach {
+            let proposal = $0.request
+            proposalPayloadsStore.set($0, forKey: proposal.proposer.publicKey)
```
flypaper0, line 44: why we cannot remove proposalPayloadsStore and depend on rpc history only? 

If big array of expired proposals the only reason I think it's fine. We can fix it when implement expiration logic for proposals

PR #751, opened 2023-03-16
[Web3Wallet] Echo

alexander-lsvk:
## Due Dilligence

* [ ] Breaking change


flypaper0 (approved):

Sources/Web3Wallet/Web3Wallet.swift
```diff
@@ -35,10 +36,11 @@ public class Web3Wallet {
     /// - Parameters:
     ///   - metadata: App metadata
     ///   - signerFactory: Auth signers factory
-    static public func configure(metadata: AppMetadata, signerFactory: SignerFactory) {
+    static public func configure(metadata: AppMetadata, signerFactory: SignerFactory, echoHost: String = "echo.walletconnect.com", environment: APNSEnvironment) {
```
llbartekll, line 39: should the echo host be an optional? not every project is using our PN solution.

llbartekll (approved):
looks good!

Sources/Web3Wallet/Web3Wallet.swift
llbartekll: add echo config? that is optional and contains both?

llbartekll: or it's good, just add a default environment?

PR #761, opened 2023-03-24
Comply with CAIP10 regex

nerzh:
# Description

Describe the bug
The address in the blockchain that we would like to support via WalletConnect contains a :
(e.g. 0:38d4c6ef92bd7ff4426eb232af3ff87bea525eff112f93eb28069e54bca89e4b),
so your conformsToCAIP10 function will not pass this address because regex is not support CAIP-10: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md#syntax

To avoid this problem, we replaced : with %3A (encoding from the ASCII table), but then your regular will not resolve %.

Can you please tell me what manual you used to write this regular? What I can see from the CAIP description in the Syntax section regarding account_id- I see that their regular supports the % symbol - [-.%a-zA-Z0-9]{1,128}.

Thanks in advance for your attention!

### DIFF
--- a/Sources/WalletConnectUtils/Extensions/String+CAIPs.swift
+++ b/Sources/WalletConnectUtils/Extensions/String+CAIPs.swift
@@ -4,7 +4,7 @@ extension String {
 
     static let chainNamespaceRegex = "^[-a-z0-9]{3,8}$"
     static let chainReferenceRegex = "^[-a-zA-Z0-9_]{1,32}$"
-    static let accountAddressRegex = "^[a-zA-Z0-9]{1,64}$"
+    static let accountAddressRegex = "^[-.%a-zA-Z0-9]{1,128}$"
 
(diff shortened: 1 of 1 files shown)


Elyniss:
Yo swift team, code checks out. I had the same change on Kotlin a day ago

nerzh:
🙄

flypaper0:
Thanks for fix. @nerzh create PR to develop branch please

flypaper0:
@nerzh created https://github.com/WalletConnect/WalletConnectSwiftV2/pull/764

nerzh:
> Thanks for fix. @nerzh create PR to develop branch please

ok, I created

ilyar:
> @nerzh created #764

with the current approach of accepting requests, it makes a very world with gosling kittens

Issue #772, opened 2023-03-31
Sample app PN feedback


chadyj:
Here is a collection of feedback from testing push in the sample app.

Some improvements or bug fixes will help dogfooding.

## PN's

When tapping a PN the app opens. Instead, I should be deep linked to the actual message

## Subscriptions List

The list seems to be un-sorted.  Sort the list by date received

Show the last received message date.
Eg GM dapp - Yesterday

Bug-  I see multiple duplicate subscriptions with empty PN's. Do we want to show these? What's the user value?

<img width="300" src="https://user-images.githubusercontent.com/116099/229051894-185d5970-5aa9-47bc-9d15-535ce530788b.jpeg">

## Notification List

Show the date

Icon's take several seconds to load.

New notifications do not show up in the list. I have to leave the screen, then come back.

## Notification detail

Missing the notification detail. The user should be able to tap on a notification and read full text, click on links, etc.

PR #805, opened 2023-04-14
[WalletApp] PN Screen - Display and sort by publishedAt, subscribe to PN publisher

radeknovis:
# Description
Small improvements

Resolves # (issue)
Resolves some of the feedback here
https://github.com/WalletConnect/WalletConnectSwiftV2/issues/772

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Example/WalletApp/PresentationLayer/Wallet/PushMessages/PushMessagesPresenter.swift
+++ b/Example/WalletApp/PresentationLayer/Wallet/PushMessages/PushMessagesPresenter.swift
@@ -7,6 +7,7 @@ final class PushMessagesPresenter: ObservableObject {
     private let interactor: PushMessagesInteractor
     private let router: PushMessagesRouter
     private var disposeBag = Set<AnyCancellable>()
+    
     @Published var pushMessages: [PushMessageViewModel] = []
 
     init(interactor: PushMessagesInteractor, router: PushMessagesRouter) {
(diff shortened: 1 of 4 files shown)


Example/WalletApp/PresentationLayer/Wallet/PushMessages/PushMessagesPresenter.swift
```diff
@@ -40,9 +41,28 @@ extension PushMessagesPresenter: SceneViewModel {
...
+            }
+            .map { pushMessageRecord in
+                PushMessageViewModel(pushMessageRecord: pushMessageRecord)
+            }
+        
+        interactor.pushMessagePublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] newPushMessage in
+                let newMessageViewModel = PushMessageViewModel(pushMessageRecord: newPushMessage)
+                guard let index = self?.pushMessages.firstIndex(
```
flypaper0, line 57: may be just update pushMessages without sorting: 
`self.pushMessages = interactor.getPushMessages()`

and create computed property in Presenter: 
```
var viewModels: [PushMessageViewModel] {
return pushMessages
.sort { ... }
.map { PushMessageViewModel(...) }
}
```

llbartekll (approved):

Example/WalletApp/PresentationLayer/Wallet/PushMessages/PushMessagesPresenter.swift
radeknovis: I would agree if there were more places where we mutate the `pushMessages` array, but we have only a single place atm; also, it's good to be cautious with computed properties used in SwiftUI Views as they can be recomputed quite often. Again it's not the problem in this case, but more like personal preference. WDYT?

PR #811, opened 2023-04-17
[CICD]  test-without-building

radeknovis:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -25,10 +25,7 @@ runs:
 
     - name: Build for testing
       shell: bash
-      env:
-        RELAY_ENDPOINT: ${{ inputs.relay-endpoint }}
-        PROJECT_ID: ${{ inputs.project-id }}
-      run: make build_all RELAY_HOST=$RELAY_ENDPOINT PROJECT_ID=$PROJECT_ID
+      run: make build_all RELAY_HOST=${{ inputs.project-id }} PROJECT_ID=${{ inputs.project-id }}
 
     - name: Tar DerivedDataCache
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
 
     - name: Untar DerivedDataCache
       shell: bash
-      run: tar xPpf products.tar
+      run: test -f products.tar && tar xPpf products.tar || echo "No artifacts to untar"
 
     # Package Unit tests
     - name: Run tests
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ build_wallet:
 	fastlane build scheme:WalletApp
 
 echo_ui_tests:
-	fastlane tests scheme:EchoUITests relay_host:$(RELAY_HOST) project_id:$(PROJECT_ID)
+	echo "EchoUITests disabled"
 
 ui_tests:
 	echo "UI Tests disabled"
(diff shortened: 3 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -14,35 +14,60 @@ concurrency:
...
+          project-id: ${{ secrets.PROJECT_ID }}
+
+  test:    
+    needs: prepare
     runs-on: macos-12
     strategy:
+      fail-fast: false
       matrix:
-        test-type: [unit-tests, integration-tests, build-example-wallet, build-example-dapp, relay-tests]
+        test-type: [integration-tests, smoke-tests, relay-tests]
```
llbartekll, line 32: smoke tests does not have to run together with integration tests.
smoke-tests is just a test plan with selected integration tests
we want to run all integration tests on our CI and smoke tests in rust workflow


why build-example-wallet, build-example-dapp are excluded now?
we need to build apps to check if sdk did not break integration

Makefile
```diff
@@ -25,29 +34,32 @@ build_wallet:
 ui_tests:
 	echo "UI Tests disabled"
 
-unit_tests:
```
llbartekll, line 28: why is it removed?

.github/workflows/ci.yml
llbartekll: I think we should run different workflows on pr to main and pr to develop

llbartekll: btw, rust is running against our release, so smoke tests can be build for testing on pr to main only

.github/workflows/ci.yml
```diff
@@ -14,35 +14,60 @@ concurrency:
...
+
+      - uses: ./.github/actions/build
+        with:
+          project-id: ${{ secrets.PROJECT_ID }}
+
+  test:    
+    needs: prepare
     runs-on: macos-12
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
radeknovis: Without this, if one job in the matrix strategy fails, it will cancel the rest of the builds. I thought it might be valuable to get back results from all the tests job.

.github/workflows/ci.yml
radeknovis: Oh right, We don't need to run both smoke_tests and integration_tests as smoke_tests are a subset.

radeknovis: for the build-example-wallet and build-example-dapp, I think this can now be verified in the prepare step when calling `build-for-testing`. I will double-check what's more efficient, having it built as part of that or having individual jobs like it was before.

Makefile
radeknovis: My bad I thought it was the same thing as integration_test. I will put it back.

.github/workflows/release.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       with:
         path: |
           .build
-          SourcePackagesCache
+          ../SourcePackagesCache
```
flypaper0, line 26: as I see SourcePackagesCache is placed in project root folder, why `../` was added?

.github/actions/build/action.yml
```diff
@@ -0,0 +1,40 @@
...
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v3
+    
+    - uses: actions/cache@v3
+      with:
+        path: |
+          SourcePackagesCache
+        key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
```
flypaper0, line 21: We using `actions/cache@v3` for caching DerivedData now (in same way with SourcePackagesCache). Why saving DerivedData separately in .tar archives is better?

.github/actions/smoke_tests/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/cache/restore@v3
+      with:
+        path: |
+          products.tar
+        key: ${{ runner.os }}-deriveddata-main-${{ inputs.commit-sha }}
```
flypaper0, line 22: Do we need to restore `SourcePackagesCache` for smoke_tests?

.github/workflows/ci.yml
radeknovis: @llbartekll can you elaborate on this? Do you mean running only specific tests when targeting develop and different tests when targeting main? 

> I think we should run different workflows on pr to main and pr to develop

.github/actions/smoke_tests/action.yml
radeknovis: In case of cache hit we should have DerivedData containing all products necessary to just execute tests with `test-without-building` so we don't need to `SourcePackagesCache`

.github/workflows/ci.yml
llbartekll: I mean that Rust relay needs to verify it's staging works with our prod before they can deploy.
and they run our smoke tests, so the build for testing(smoke_tests) artifact needs to reflect our prod, which is main branch.
So we don't need to build for testing on develop merges as it does not reflect our prod.

llbartekll (approved):

flypaper0 (approved):

PR #814, opened 2023-04-20
Move `fatalError()` call before `return` otherwise it never gets called

hboon:
This also removes the build warning:

> Code after 'return' will never be executed

The pre-PR state is only correct if the `fatalError()` call was only meant to be a marked to be grepped and not intended to crash at dev/runtime


radeknovis:
It will be addressed as part of the next PR. Thanks for raising this.

PR #820, opened 2023-04-26
[W3W] Verify API

alexander-lsvk:
## Due Dilligence

* [ ] Breaking change


Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
```diff
@@ -1,41 +1,61 @@
...
                 )
-
+                
                 let request = AuthRequest(id: payload.id, payload: payload.request.payloadParams)
-                onRequest?(request)
+                
+                if let rawRequest = payload.rawRequest, let verifyClient {
+                    Task(priority: .high) {
+                        let attestationId = rawRequest.rawRepresentation.sha256().toHexString()
+                        let origin = try? await verifyClient.registerAssertion(attestationId: attestationId)
```
llbartekll, line 48: why wallet registers assertion here?
afaik, wallet should only verify origin

Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
```diff
@@ -277,7 +282,21 @@ private extension ApproveEngine {
             return respondError(payload: payload, reason: .invalidUpdateRequest, protocolMethod: SessionProposeProtocolMethod())
         }
         proposalPayloadsStore.set(payload, forKey: proposal.proposer.publicKey)
-        onSessionProposal?(proposal.publicRepresentation(pairingTopic: payload.topic))
+        
+        if let rawRequest = payload.rawRequest, let verifyClient {
+            Task(priority: .high) {
+            let attestationId = rawRequest.rawRepresentation.sha256().toHexString()
+                let origin = try? await verifyClient.registerAssertion(attestationId: attestationId)
```
llbartekll, line 289: same here, shouldn't you call verifyOrigin() method on verify client?
I can see it's not called anywhere 🤔

Package.swift
```diff
@@ -111,10 +111,10 @@ let package = Package(
             dependencies: []),
         .target(
             name: "WalletConnectVerify",
-            dependencies: ["WalletConnectUtils"]),
+            dependencies: ["WalletConnectUtils", "WalletConnectNetworking"]),
```
llbartekll, line 114: probably would make sense in the future to extract HTTP from WalletConnectNetworking to not make verify dependent to the whole networking

Package.swift
```diff
@@ -48,7 +48,7 @@ let package = Package(
     targets: [
         .target(
             name: "WalletConnectSign",
-            dependencies: ["WalletConnectPairing"],
+            dependencies: ["WalletConnectPairing", "WalletConnectVerify"],
```
llbartekll, line 51: As discussed I would not make WalletConnectSign dependent from WalletConnectVerify
but WalletConnectNetworking instead

Sources/Auth/Types/Public/AuthContext.swift
```diff
@@ -0,0 +1,11 @@
+public struct AuthContext: Equatable, Hashable {
```
llbartekll, line 1: VerifyContext would be shared between Auth and Sign

Sources/WalletConnectKMS/Serialiser/Serializer.swift
```diff
@@ -40,12 +40,13 @@ public class Serializer: Serializing {
...
     /// - Returns: Deserialized object
-    public func deserialize<T: Codable>(topic: String, encodedEnvelope: String) throws -> (T, derivedTopic: String?) {
+    public func deserialize<T: Codable>(topic: String, encodedEnvelope: String) throws -> (T, derivedTopic: String?, rawData: String?) {
         let envelope = try Envelope(encodedEnvelope)
         switch envelope.type {
         case .type0:
             let deserialisedType: T = try handleType0Envelope(topic, envelope)
-            return (deserialisedType, nil)
+            let rawRequest = try rawRequest(topic, envelope)
+            return (deserialisedType, nil, rawRequest)
```
llbartekll, line 49: would make sense that `handleType0Envelope` returns the rawRequest too? like `(T, Data)`
we would do the decryption computation(`codec.decode`) only once then.
now we do it twice: first in `handleType0Envelope` function and then in `try rawRequest(topic, envelope)`

can we also rename `rawRequest`? as it can be a rpc request and response and in theory even more.
just `raw`, `rawDecrypted` `rawDecryptedPayload`?

Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
```diff
@@ -235,8 +242,21 @@ private extension SessionEngine {
...
-        onSessionRequest?(request)
+        
+        if let rawRequest = payload.rawRequest, let verifyClient {
+            Task(priority: .high) {
+                let attestationId = rawRequest.rawRepresentation.sha256().toHexString()
+                let origin = try? await verifyClient.registerAssertion(attestationId: attestationId)
+                let sessionContext = await Session.Context(
+                    origin: origin,
+                    validation: (origin == session.peerParticipant.metadata.url) ? .valid : (origin == nil ? .unknown : .invalid),
+                    verifyUrl: verifyClient.verifyHost
```
llbartekll, line 253: same here. `verifyOrigin` and I would handle it generically through Networking package

Sources/WalletConnectVerify/AssertionRegistrer.swift
```diff
@@ -1,5 +1,26 @@
...
+    func registerAssertion(attestationId: String) async throws -> String {
+        let httpClient = HTTPNetworkClient(host: verifyHost)
+        let response = try await httpClient.request(
+            VerifyResponse.self,
+            at: VerifyAPI.resolve(attestationId: attestationId)
+        )
+        guard let origin = response.origin else {
+            throw Errors.registrationFailed
+        }
+        return origin
```
llbartekll, line 24: as discussed, should be executed by verifyOrigin() function

Sources/WalletConnectVerify/AttestKeyGenerator.swift
```diff
@@ -21,8 +20,9 @@ class AttestKeyGenerator: AttestKeyGenerating {
     }
 
     func generateKeys() async throws -> String {
-        let keyId = try await service.generateKey()
-        keyIdStorage.set(keyId, forKey: Constants.keyIdStorageKey)
-        return keyId
+        // TODO - Add DCAppAttestService handling for iOS 14.0
+        // let keyId = try await service.generateKey()
+        keyIdStorage.set("keyId", forKey: Constants.keyIdStorageKey)
+        return "keyId"
```
llbartekll, line 26: what does it change? this class is used ably by the dapp, which we don't support natively now.

Sources/Web3Wallet/Web3WalletClient.swift
```diff
@@ -12,21 +12,21 @@ public class Web3WalletClient {
...
-    public var sessionRequestPublisher: AnyPublisher<Request, Never> {
+    public var sessionRequestPublisher: AnyPublisher<(request: Request, context: Session.Context?), Never> {
         signClient.sessionRequestPublisher.eraseToAnyPublisher()
     }
     
     /// Publisher that sends authentication requests
     ///
     /// Wallet should subscribe on events in order to receive auth requests.
-    public var authRequestPublisher: AnyPublisher<AuthRequest, Never> {
+    public var authRequestPublisher: AnyPublisher<(request: AuthRequest, context: AuthContext?), Never> {
```
llbartekll, line 29: Session.Context and AuthContext, kind of inconsistent

Sources/WalletConnectKMS/Serialiser/Serializer.swift
```diff
@@ -59,15 +60,23 @@ public class Serializer: Serializing {
         }
     }
 
-    private func handleType1Envelope<T: Codable>(_ topic: String, peerPubKey: Data, sealbox: Data) throws -> (T, String) {
+    private func rawRequest(_ topic: String, _ envelope: Envelope) throws -> String? {
+        if let symmetricKey = kms.getSymmetricKeyRepresentable(for: topic) {
+            return try String(data: codec.decode(sealbox: envelope.sealbox, symmetricKey: symmetricKey), encoding: .utf8)
```
llbartekll, line 65: why do you convert it into the string as later you hash rawRepresentation of that string?

llbartekll (changes requested):

Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
llbartekll: So I would add `verifiable` property to `ProtocolMethod`, call `verifyOrigin` in NetworkingInteractor and pass VerifyContext to requestSubscription

@flypaper0 what do you think?

Example/WalletApp/PresentationLayer/Wallet/AuthRequest/AuthRequestPresenter.swift
```diff
@@ -18,12 +19,14 @@ final class AuthRequestPresenter: ObservableObject {
...
         router: AuthRequestRouter,
-        request: AuthRequest
+        request: AuthRequest,
+        context: AuthContext?
     ) {
         defer { setupInitialState() }
         self.interactor = interactor
         self.router = router
         self.request = request
+        self.verified = (context?.validation == .valid) ? true : (context?.validation == .unknown ? nil : false)
```
flypaper0, line 29: mb we can add computed property `isValid: Bool` to Context ?

Sources/Auth/AuthClientFactory.swift
```diff
@@ -44,11 +46,12 @@ public struct AuthClientFactory {
...
+        let verifyClient = try? VerifyClientFactory.create(verifyHost: metadata.verifyUrl)
         let messageVerifierFactory = MessageVerifierFactory(crypto: crypto)
         let signatureVerifier = messageVerifierFactory.create(projectId: projectId)
         let appRespondSubscriber = AppRespondSubscriber(networkingInteractor: networkingClient, logger: logger, rpcHistory: history, signatureVerifier: signatureVerifier, pairingRegisterer: pairingRegisterer, messageFormatter: messageFormatter)
         let walletErrorResponder = WalletErrorResponder(networkingInteractor: networkingClient, logger: logger, kms: kms, rpcHistory: history)
-        let walletRequestSubscriber = WalletRequestSubscriber(networkingInteractor: networkingClient, logger: logger, kms: kms, walletErrorResponder: walletErrorResponder, pairingRegisterer: pairingRegisterer)
+        let walletRequestSubscriber = WalletRequestSubscriber(networkingInteractor: networkingClient, logger: logger, kms: kms, walletErrorResponder: walletErrorResponder, pairingRegisterer: pairingRegisterer, verifyClient: verifyClient)
```
flypaper0, line 54: VerifyClient could be nil right? What does it mean if VerifyClient is nil?

Sources/Auth/Services/Wallet/WalletRequestSubscriber.swift
```diff
@@ -1,41 +1,61 @@
...
                 pairingRegisterer.activate(
                     pairingTopic: payload.topic,
                     peerMetadata: payload.request.requester.metadata
                 )
-
+                
                 let request = AuthRequest(id: payload.id, payload: payload.request.payloadParams)
-                onRequest?(request)
+                
+                if let rawRequest = payload.rawRequest, let verifyClient {
```
flypaper0, line 45: I see the same validation code it SessionEngine, is it possible to do it in one-two lines? May be hide it in some kind of ValidationService and reuse

Sources/WalletConnectKMS/Serialiser/Serializer.swift
```diff
@@ -59,15 +60,23 @@ public class Serializer: Serializing {
         }
     }
 
-    private func handleType1Envelope<T: Codable>(_ topic: String, peerPubKey: Data, sealbox: Data) throws -> (T, String) {
+    private func rawRequest(_ topic: String, _ envelope: Envelope) throws -> String? {
+        if let symmetricKey = kms.getSymmetricKeyRepresentable(for: topic) {
```
flypaper0, line 64: guard?

Sources/WalletConnectKMS/Serialiser/Serializer.swift
```diff
@@ -59,15 +60,23 @@ public class Serializer: Serializing {
...
+    private func handleType1Envelope<T: Codable>(_ topic: String, peerPubKey: Data, sealbox: Data) throws -> (T, String, String?) {
         guard let selfPubKey = kms.getPublicKey(for: topic)
         else { throw Errors.publicKeyForTopicNotFound }
 
         let agreementKeys = try kms.performKeyAgreement(selfPublicKey: selfPubKey, peerPublicKey: peerPubKey.toHexString())
         let decodedType: T = try decode(sealbox: sealbox, symmetricKey: agreementKeys.sharedKey.rawRepresentation)
         let derivedTopic = agreementKeys.derivedTopic()
         try kms.setAgreementSecret(agreementKeys, topic: derivedTopic)
-        return (decodedType, derivedTopic)
+        return (decodedType, derivedTopic, nil)
```
flypaper0, line 79: Why we return nil rawRequest here?

Sources/WalletConnectKMS/Serialiser/Serializer.swift
flypaper0: agree with @llbartekll 
if we rename it to rawPayload we could make it not optional, and return it for both type1 and type0 envelopes

Sources/WalletConnectVerify/Register/VerifyService.swift
```diff
@@ -0,0 +1,34 @@
+import Foundation
+import WalletConnectNetworking
+
+enum VerifyAPI: HTTPService {
```
llbartekll, line 4: probably for this single request would be easier to use raw urlsession

Sources/WalletConnectVerify/VerifyClient.swift
```diff
@@ -27,12 +31,19 @@ public actor VerifyClient {
         try await appAttestationRegistrer.registerAttestationIfNeeded()
     }
 
-    public func verifyOrigin() async throws {
-        try await originVerifier.verifyOrigin()
+    public func verifyOrigin(assertionId: String) async throws -> String {
+        return try await originVerifier.verifyOrigin(assertionId: assertionId)
+    }
+    
+    public func createVerifyContext(origin: String?, domain: String) -> VerifyContext {
```
llbartekll, line 38: 👍

Sources/WalletConnectVerify/VerifyClientFactory.swift
```diff
@@ -1,24 +1,31 @@
...
-@available(iOS 14.0, *)
-@available(macOS 11.0, *)
 public class VerifyClientFactory {
-
-    public static func create() throws -> VerifyClient {
-        let originVerifier = OriginVerifier()
+    public static func create(verifyHost: String?) throws -> VerifyClient? {
+        guard let verifyHost else {
+            return nil
+        }
```
llbartekll, line 8: wouldn't be better to just do like?:
```suggestion
    public static func create(verifyHost: String) throws -> VerifyClient? {
```

llbartekll (approved):
looks good, just few comments

PR #853, opened 2023-05-18
Feature/w3m networking

radeknovis:
# Description

- Adds networking layer
- Adds fetching for ios supported wallets


Sources/Web3Modal/Environment+ProjectId.swift
```diff
@@ -0,0 +1,12 @@
+import SwiftUI
+
+private struct ProjectIdKey: EnvironmentKey {
```
flypaper0, line 3: looks nice, where it's using?

flypaper0 (approved):
looks good

Sources/Web3Modal/Environment+ProjectId.swift
radeknovis: Yeah I thought about using it to set projectID but didn't end up using it

PR #860, opened 2023-05-22
Fix SPM dependency on QRCode

radeknovis:


llbartekll (approved):
did you try if it resolves?

flypaper0 (approved):

PR #869, opened 2023-05-24
[DApp, WalletApp] add uri encoding

maltsevoff:
# Description

* added uri percent encoding in dapp. This is necessary in order to parse universal link with wallet connect pairing uri parameter in real applications. Pairing uri contains extra '?' which fails normal query parameters parsing. 
* improved pairing uri parsing in WalletApp with query parameters handling.

# Resolves (issue)

It resolves inability to send correct pairing uri via universal link on iOS. 

## How Has This Been Tested?

* you can try to connect dapp without percent encoding to any mobile wallet (e.g. TrustWallet)
* repeat it with percent encoding 

Also it would be nice to fix your documentation by [link](https://docs.walletconnect.com/2.0/ios/guides/mobile-linking) in 'iOS Wallet support' and provide correct universal link for with uri encoding.

### DIFF
--- a/Example/DApp/Sign/Connect/ConnectViewController.swift
+++ b/Example/DApp/Sign/Connect/ConnectViewController.swift
@@ -63,7 +63,8 @@ class ConnectViewController: UIViewController, UITableViewDataSource, UITableVie
     }
 
     @objc func connectWithExampleWallet() {
-        let url = URL(string: "https://walletconnect.com/wc?uri=\(uri.absoluteString)")!
+		let encodedUri = uri.absoluteString.addingPercentEncoding(withAllowedCharacters: .rfc3986)!
+        let url = URL(string: "walletapp://wc?uri=\(encodedUri)")!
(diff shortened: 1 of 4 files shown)


Example/DApp/Sign/Connect/ConnectViewController.swift
```diff
@@ -63,7 +63,8 @@ class ConnectViewController: UIViewController, UITableViewDataSource, UITableVie
     }
 
     @objc func connectWithExampleWallet() {
-        let url = URL(string: "https://walletconnect.com/wc?uri=\(uri.absoluteString)")!
+		let encodedUri = uri.absoluteString.addingPercentEncoding(withAllowedCharacters: .rfc3986)!
```
radeknovis, line 66: `WalletConnectURI` object already has a convenience computed property called`deeplinkUri` so you can use that.

maltsevoff:
@radeknovis yes, doing it

PR #875, opened 2023-05-26
[CI&CD] Fix CI runs for PRs from forks

radeknovis:
More context here: https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets

NB: This is complicated to test as it needs to be merged in `main` first, so it is split in two parts leaving original `ci` workflow unchanged

### DIFF
--- a/.github/workflows/ci_v2.yml
+++ b/.github/workflows/ci_v2.yml
@@ -1,0 +1,100 @@
+name: ci_v2
+
+on:
+  pull_request_target:
+    branches:
+      - develop
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.event.pull_request.number || github.ref_name }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request_target' }}
+    
+jobs:
+  authorize:
+    environment:
+      ${{ (github.event_name == 'pull_request_target' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_v2.yml
```diff
@@ -0,0 +1,100 @@
+name: ci_v2
+
+on:
+  pull_request_target:
```
radeknovis, line 4: Runs triggered using this event have access to secrets.

.github/workflows/ci_v2.yml
```diff
@@ -0,0 +1,100 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request_target' && github.event.pull_request.number || github.ref_name }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request_target' }}
+    
+jobs:
+  authorize:
+    environment:
+      ${{ (github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository) &&
+      'external' || 'internal' }}
```
radeknovis, line 18: This `external` vs `internal` environment would differentiate between regular PRs and PRs incoming from forks and we can require approval and we can specify that `external` ones need approval before running.

llbartekll (approved):

PR #891, opened 2023-06-01
[CI/CD] slack notification action

llbartekll:
# Description

move slack notification action to run tests without building
Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,12 +72,6 @@ jobs:
       shell: bash
       run: make relay_tests RELAY_HOST=relay.walletconnect.com PROJECT_ID=${{ secrets.PROJECT_ID }}
 
-    # Smoke tests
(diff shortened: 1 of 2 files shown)


radeknovis (approved):

.github/workflows/ci.yml
```diff
@@ -72,21 +72,6 @@ jobs:
       shell: bash
       run: make relay_tests RELAY_HOST=relay.walletconnect.com PROJECT_ID=${{ secrets.PROJECT_ID }}
 
-    # Smoke tests
-    - name: Run smoke tests
```
flypaper0, line 76: why smoke-tests removed from ci action

llbartekll: correct me if I am wrong but we don't run them in this workflow. And we have a run smoke tests action in run tests without building

PR #892, opened 2023-06-02
[Chat] History API

flypaper0:
# Description
- HistoryClient
- HistoryClient integration tests
- History integrated into chat sdk
- KeyedDatabase refactor, legacy store removed
- Messages batch update

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Example/IntegrationTests/History/HistoryTests.swift
```diff
@@ -0,0 +1,79 @@
...
+        try await historyClient.registerTags(
+            payload: RegisterPayload(tags: [String(tag)], relayUrl: relayUrl),
+            historyUrl: historyUrl)
+
+        try await relayClient2.subscribe(topic: topic)
+        try await relayClient1.publish(topic: topic, payload: payload, tag: tag, prompt: false, ttl: 3000)
+
+        wait(for: [exp], timeout: InputConfig.defaultTimeout)
+
+        sleep(5) // History server has a queue
```
llbartekll, line 67: may the test fail sometimes due to insufficient delay?

Example/IntegrationTests/History/HistoryTests.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+import Combine
+import XCTest
+@testable import WalletConnectHistory
+
+final class HistoryTests: XCTestCase {
```
llbartekll, line 6: please exclude those tests from smoke test plan

Sources/Chat/ProtocolServices/History/HistoryService.swift
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    func fetchMessageHistory(thread: Thread) async throws -> [Message] {
+        let wrappers: [MessagePayload.Wrapper] = try await historyClient.getMessages(
+            topic: thread.topic,
+            count: 200, direction: .backward
+        )
+
+        return wrappers.map { wrapper in
+            let (messagePayload, messageClaims) = try! MessagePayload.decodeAndVerify(from: wrapper)
```
llbartekll, line 24: should we force unwrap here?
would make sense to compactMap nil values or throw an error instead?

llbartekll (approved):

Example/IntegrationTests/History/HistoryTests.swift
flypaper0: removed from smoke tests

PR #896, opened 2023-06-05
remove client id from clients config

llbartekll:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


flypaper0 (approved):
looks good. I think it's enough to fix crash

PR #898, opened 2023-06-07
[W3M] SDK Interface cleanup

radeknovis:


Sources/Web3Modal/Web3ModalClient.swift
```diff
@@ -0,0 +1,160 @@
...
+            requiredNamespaces: Web3Modal.config.sessionParams.requiredNamespaces,
+            optionalNamespaces: Web3Modal.config.sessionParams.optionalNamespaces,
+            sessionProperties: Web3Modal.config.sessionParams.sessionProperties,
+            topic: topic
+        )
+    }
+    
+    /// For proposing a session to a wallet.
+    /// Function will propose a session on newly created pairing.
+    public func createPairingAndConnect() async throws -> WalletConnectURI {
```
llbartekll, line 82: would make sense to call the method just connect() with an optional topic?
same as deprecated method of sign client

```
    public func connect(
        requiredNamespaces: [String: ProposalNamespace],
        optionalNamespaces: [String: ProposalNamespace]? = nil,
        sessionProperties: [String: String]? = nil,
        topic: String? = nil
```

llbartekll: actually would match kotlin's interface too

Sources/Web3Modal/Web3Modal.swift
```diff
@@ -0,0 +1,126 @@
...
+        metadata: AppMetadata,
+        sessionParams: SessionParams = .default
+    ) {
+        Pair.configure(metadata: metadata)
+        Web3Modal.config = Web3Modal.Config(projectId: projectId, sessionParams: sessionParams)
+    }
+    
+    public static func set(sessionParams: SessionParams) {
+        Web3Modal.config.sessionParams = sessionParams
+    }
```
llbartekll, line 56: what the set method is for?
what's the difference between configure()?

llbartekll (approved):

Sources/Web3Modal/Web3Modal.swift
llbartekll: I see, does it work? It’s singletons and config is used on the instantiation, so with the first call only

radeknovis: Yes, it works. Config is stored as a static variable, which can be changed only via these 2 methods. I'm calling `set()` from Dapp

PR #943, opened 2023-07-05
[WCM] Implement missing ui + fix ui glitches

radeknovis:


.gitignore
```diff
@@ -31,3 +31,5 @@ DerivedDataCache
 *.ipa
 *.zip
 test_results/
+
+Sources/WalletConnectModal/Secrets/secrets.json
```
flypaper0, line 35: When it used? Other team also need to have this file locally?

Package.swift
```diff
@@ -132,7 +132,10 @@ let package = Package(
             dependencies: ["WalletConnectUtils", "WalletConnectNetworking"]),
         .target(
             name: "WalletConnectModal",
-            dependencies: ["QRCode", "WalletConnectSign"]),
+            dependencies: ["QRCode", "WalletConnectSign"],
+            exclude: ["Secrets/secrets.json.sample"],
+            resources: [.copy("Secrets/secrets.json")]
```
flypaper0, line 137: Strange. You're ignoring this file by git, but it's required for WalletConnectModal SDK

Sources/WalletConnectModal/Modal/Modal+Previews.swift
```diff
@@ -34,7 +33,7 @@ struct ModalContainerView_Previews: PreviewProvider {
         
         init() {
             
-            let projectId = "9bfe94c9cbf74aaa0597094ef561f703"
+            let projectId = Secrets.load().projectID
```
flypaper0, line 36: Now we're taking Project ID from configuration. Why json file is better?

Sources/WalletConnectModal/Networking/Explorer/ListingsResponse.swift
```diff
@@ -34,3 +34,15 @@ struct Listing: Codable, Hashable, Identifiable {
         let universal: String?
     }
 }
+
+#if DEBUG
```
flypaper0, line 38: I'd move all stubs in separate folder "Stubs". To not pollute SDK code

Sources/WalletConnectModal/UI/Common/AsyncImage.swift
```diff
@@ -12,7 +12,7 @@ struct AsyncImage<Content>: View where Content: View {
             
             var request = URLRequest(url: url)
             request.setValue(ExplorerAPI.userAgent, forHTTPHeaderField: "User-Agent")
-            request.setValue(WalletConnectModal.config.metadata.name, forHTTPHeaderField: "Referer")
+            request.setValue(WalletConnectModal.config?.metadata.name, forHTTPHeaderField: "Referer")
```
flypaper0, line 15: Could you double check that it sets normal string instead of `Optional("...")`

Sources/WalletConnectModal/Modal/Modal+Previews.swift
radeknovis: `configuration.xcconfig` is not available from package, only for stuff in the ExampleApp project, or at least I haven't figured out how to read from it.

I'm not too keen on having two approaches how to share secrets/vars either. So any suggestions on how to do this better are more than welcome :D

Package.swift
radeknovis: It is ignored, so we don't commit actual secrets. This will pass if the file is not present, but to access the JSON with actual secrets, you need to bundle it with the target.

.gitignore
radeknovis: Only needed if you want to run WalletConnectModal previews. If you don't have it, you will get a somewhat explanatory error to create it.

Sources/WalletConnectModal/UI/Common/AsyncImage.swift
radeknovis: There is a unit test for this, verifying that both `User-Agent` and `Referer` headers are correctly populated.

radeknovis:
@flypaper0 I've tried to address all your comments. Can you have a look again, please?

Sources/WalletConnectModal/Modal/Modal+Previews.swift
llbartekll: you can pass projectID to w3m in a config method like in `Relay `

```
    static public func configure(
        relayHost: String = "relay.walletconnect.com",
        projectId: String,
        socketFactory: WebSocketFactory,
        socketConnectionType: SocketConnectionType = .automatic
    )
```

llbartekll: can't believe we missed comiting proj id xd

radeknovis: Yeah, this is how it will be done when actually integrating the modal SDK https://github.com/WalletConnect/WalletConnectSwiftV2/blob/9cead7d4443815687e02cd87ba888130225283fc/Sources/WalletConnectModal/WalletConnectModal.swift#L51-L57C8

but this is in the context of previews within SDK itself so I don't have any projectID injected from elsewhere.

llbartekll (approved):

PR #950, opened 2023-07-07
[Push] HistoryAPI integration. Cold start

flypaper0:


Sources/WalletConnectPush/Client/Dapp/DappPushClientFactory.swift
```diff
@@ -19,9 +19,9 @@ public struct DappPushClientFactory {
     static func create(metadata: AppMetadata, logger: ConsoleLogging, keyValueStorage: KeyValueStorage, keychainStorage: KeychainStorageProtocol, networkInteractor: NetworkInteracting, syncClient: SyncClient) -> DappPushClient {
         let kms = KeyManagementService(keychain: keychainStorage)
         let subscriptionStore: SyncStore<PushSubscription> = SyncStoreFactory.create(name: PushStorageIdntifiers.pushSubscription, syncClient: syncClient, storage: keyValueStorage)
-
+        let messagesStore = KeyedDatabase<PushMessageRecord>(storage: keyValueStorage, identifier: PushStorageIdntifiers.pushMessagesRecords)
         let subscriptionStoreDelegate = PushSubscriptionStoreDelegate(networkingInteractor: networkInteractor, kms: kms)
-        let pushStorage = PushStorage(subscriptionStore: subscriptionStore, subscriptionStoreDelegate: subscriptionStoreDelegate)
+        let pushStorage = PushStorage(subscriptionStore: subscriptionStore, messagesStore: messagesStore, subscriptionStoreDelegate: subscriptionStoreDelegate)
```
llbartekll, line 24: actually dapp does not require sync as it will be deprecated and it's used only for integration tests but as you did it already we can leave it ;)

Sources/WalletConnectPush/Client/Wallet/PushSyncService.swift
```diff
@@ -23,4 +41,72 @@ final class PushSyncService {
...
+                count: 200,
+                direction: .backward
+            )
+
+            let messageRecords = historyRecords.map { record in
+                return PushMessageRecord(
+                    id: record.id.string,
+                    topic: subscription.topic,
+                    message: record.object,
+                    publishedAt: Date()
```
llbartekll, line 92: can we get a date from a history somehow? in w3i there is a label that displays the date. With this implementation the date will be incorrect.

llbartekll (commented):
when the PR is ready just ping me so I can test in the app.

PR #969, opened 2023-07-18
[WCM] Landscape improvements

radeknovis:


Sources/WalletConnectModal/Modal/ModalSheet.swift
```diff
@@ -81,14 +88,31 @@ public struct ModalSheet: View {
...
+                    .padding(.vertical, 4)
+                    .padding(.horizontal, 10)
+                    .background(Color.background3)
+                    .foregroundColor(searchEditing ? .foreground1 : .foreground3)
+                    .cornerRadius(28)
+                    .overlay(
+                        RoundedRectangle(cornerRadius: 28)
+                            .stroke(searchEditing ? Color.accent : Color.thinOverlay, lineWidth: 1)
+                    )
+                    .onDisappear {
```
flypaper0, line 107: Can we move it on parent view from `body` ?

Sources/WalletConnectModal/Modal/ModalSheet.swift
```diff
@@ -81,14 +88,31 @@ public struct ModalSheet: View {
...
+                        searchEditing = false
+                    }
+                    .transform {
+                        #if os(iOS)
+                            $0.autocapitalization(.none)
```
flypaper0, line 112: why autocapitalization is inside transform?

Sources/WalletConnectModal/Modal/Screens/WalletList.swift
```diff
@@ -116,20 +118,22 @@ struct WalletList: View {
     func viewAllItem() -> some View {
         VStack {
             VStack(spacing: 3) {
-                let startingIndex = (2 * numberOfColumns - 1)
+                let viewAllWalletsFirstRow = wallets.dropFirst(2 * numberOfColumns - 1).prefix(2)
```
flypaper0, line 121: Is it safe? prefix is crashing app if out of index as I remember

Sources/WalletConnectModal/Modal/Screens/WhatIsWalletView.swift
```diff
@@ -1,29 +1,40 @@
...
-            )
-            HelpSection(
-                title: "Your gateway to a new web",
-                description: "With your wallet, you can explore and interact with DeFi, NFTs, DAOs, and much more.",
-                assets: [.Browser, .Noun, .DAO]
-            )
-            
+        content()
+            .onAppear {
+                UIPageControl.appearance().currentPageIndicatorTintColor = AssetColor.foreground1.uiColor
```
flypaper0, line 12: AppearanceConfigurator could use for setting up appearance()

Sources/WalletConnectModal/Modal/ModalSheet.swift
radeknovis: It's related only to `TextField`s  onDisappear , otherwise I would agree with you to put these lifecycle modifier to top most view

Sources/WalletConnectModal/Modal/ModalSheet.swift
radeknovis: For some reason, `autocapitalization` is not available on macOS, so I had to wrap it in this `#if os()` and this `transform` helper method allows to do that inline rather than having the entire view duplicated with `if/else`.

However I've change the condition to really reflect the problem, and add `#if not macOS`

radeknovis: The transform method is really this simple
```swift
func transform(@ViewBuilder _ transform: (Self) -> some View) -> some View {
        transform(self)
    }
```

Sources/WalletConnectModal/UI/Common/Toast.swift
```diff
@@ -1,10 +1,10 @@
 import SwiftUI
 
 struct Toast: Equatable {
-  var style: ToastStyle
-  var message: String
-  var duration: Double = 3
-  var width: Double = .infinity
+    var style: ToastStyle
```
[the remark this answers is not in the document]
radeknovis: good spot, changed to `let`

Sources/WalletConnectModal/Modal/Screens/WalletList.swift
radeknovis: It's crashing when the value to prefix ends up being negative, it's okay if you prefix higher number of elements than there is in the array. So in this case it's safe.

Sources/WalletConnectModal/Modal/Screens/WhatIsWalletView.swift
radeknovis: We have AppearanceConfigurator for the sample apps, but not in the modal. This is the only place where we deal with appearance so far. Anyway, I learned that we will end up removing this screen from WalletConnect Modal, and it will be only in Web3Modal with a slightly different design, so I'm tempted to keep it in and deal with it properly when we do Web3Modal design

flypaper0 (approved):
lgtm

PR #976, opened 2023-07-24
[WCM] Wallet loading screen

radeknovis:

### DIFF
--- a/Sources/WalletConnectModal/Modal/Screens/GetAWalletView.swift
+++ b/Sources/WalletConnectModal/Modal/Screens/GetAWalletView.swift
@@ -31,7 +31,6 @@ struct GetAWalletView: View {
             .frame(minHeight: 400)
             .listStyle(.plain)
             
-            
             VStack(alignment: .center, spacing: 8) {
                 Text("Not what you’re looking for?")
                     .font(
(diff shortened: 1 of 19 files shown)


Sources/WalletConnectModal/Modal/Screens/GetAWalletView.swift
```diff
@@ -67,6 +66,8 @@ struct GetAWalletView: View {
     }
 }
 
+#if DEBUG
```
flypaper0, line 69: Could you please double check that previews is wrapped in DEBUG everywhere?

Sources/WalletConnectModal/Modal/Screens/WalletList.swift
```diff
@@ -188,50 +185,15 @@ struct WalletList: View {
                 $0.onTapGesture {
                     withAnimation {
                         navigateTo(.walletDetail(wallet))
+                        
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
flypaper0, line 189: why we need delay?

flypaper0 (approved):

Sources/WalletConnectModal/Modal/Screens/WalletList.swift
radeknovis: for slightly better UX, users will see `Continue in WalletXYZ` briefly and then it actually performs the deeplink

radeknovis: left comment explaining this

Sources/WalletConnectModal/Modal/Screens/GetAWalletView.swift
radeknovis: Added where it was missing

PR #991, opened 2023-07-27
[WCM] Detect installed wallets

radeknovis:


Example/DApp/Info.plist
```diff
@@ -2,6 +2,15 @@
...
 <dict>
+	<key>LSApplicationQueriesSchemes</key>
+	<array>
+		<string>metamask</string>
+		<string>trust</string>
+		<string>safe</string>
+		<string>zerion</string>
+		<string>rainbow</string>
+		<string>spot</string>
+	</array>
```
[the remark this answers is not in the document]
radeknovis: This is probably a great way how to deal with it forward, but for now, we are sticking to simple solution and mentioning it in docs that in order to detect installed wallets they need to populate these `LSApplicationQueriesSchemes`

Sources/WalletConnectModal/Modal/ModalViewModel.swift
```diff
@@ -221,6 +232,23 @@ private extension ModalViewModel {
...
+                let walletScheme = $0.mobile.native,
+                !walletScheme.isEmpty,
+                schemes.contains(walletScheme.replacingOccurrences(of: "://", with: ""))
+            {
+                $0.installed = uiApplicationWrapper.canOpenURL(URL(string: walletScheme)!)
+            }
+        }
+    }
+    
     func loadRecentWallets() {
```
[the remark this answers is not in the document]
radeknovis: I moved some of the logic to the `RecentWalletsStorage` object and tidied up the sorting logic to be more declarative.

Sources/WalletConnectModal/Modal/ModalViewModel.swift
```diff
@@ -164,79 +170,104 @@ final class ModalViewModel: ObservableObject {
...
+            {
+                $0.installed = uiApplicationWrapper.canOpenURL(URL(string: walletScheme)!)
+            }
+        }
+    }
+    
+    func loadRecentWallets() {
+        recentWalletStorage.recentWallets.forEach { wallet in
+            guard let lastTimeUsed = wallet.lastTimeUsed else { return }
+            setLastTimeUsed(wallet.id, date: lastTimeUsed)
```
flypaper0, line 262: we already have wallet instance here, why we need to search wallet by id in setLastTimeUsed again?

Sources/WalletConnectModal/Modal/RecentWalletStorage.swift
```diff
@@ -9,23 +9,44 @@ final class RecentWalletsStorage {
...
+        guard
+            let data = defaults.data(forKey: "recentWallets"),
+            let wallets = try? JSONDecoder().decode([Listing].self, from: data)
+        else {
+            return []
+        }
+        
+        return wallets.filter { listing in
+            guard let lastTimeUsed = listing.lastTimeUsed else {
+                assertionFailure("Shouldn't happen we stored wallet without `lastTimeUsed`")
```
flypaper0, line 29: Could you explain why we are storing without lastTimeUsed please?

Sources/WalletConnectModal/Modal/ModalViewModel.swift
```diff
@@ -164,79 +170,104 @@ final class ModalViewModel: ObservableObject {
...
+// MARK: - Sorting and filtering
 
-private extension ModalViewModel {
+private extension Array where Element: Listing {
     
-    func sortByRecent(_ input: [Listing]) -> [Listing] {
-        input.sorted { lhs, rhs in
-            guard let lhsLastTimeUsed = lhs.lastTimeUsed else {
+    func sortByOrder() -> [Listing] {
+        self.sorted {
```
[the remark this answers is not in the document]
radeknovis: Actually, since it is in a single computed property like this, which does not sort in place but returns a new ordered sequence, it should trigger a single UI update once it orders everything.

```swift
var filteredWallets: [Listing] {
        wallets
            .sortByOrder()
            .sortByInstalled()
            .sortByRecent()
            .filter(searchTerm: searchTerm)
}
 ```
 
 One concern might be that we are doing unnecessary iterations since `order` and `installed` won't change, so I will extract those to run only once.

File: Sources/WalletConnectModal/Modal/ModalViewModel.swift (hunk 3 above)
Comment (radeknovis) [reply Sources/WalletConnectModal/Modal/ModalViewModel.swift:262]: The wallet instance here is the one loaded from UserDefaults, while the one I'm matching in `setLastTimeUsed` is part of my `wallets` state property.

File: Sources/WalletConnectModal/Modal/RecentWalletStorage.swift (hunk 4 above)
Comment (radeknovis) [reply Sources/WalletConnectModal/Modal/RecentWalletStorage.swift:29]: It should never happen that we store it without `lastTimeUsed`, so I've added this just as an extra check If we ever do.

File: Example/DApp/Info.plist [hunk 6]
```diff
@@ -2,6 +2,15 @@
...
 <plist version="1.0">
 <dict>
+	<key>LSApplicationQueriesSchemes</key>
+	<array>
+		<string>metamask</string>
+		<string>trust</string>
+		<string>safe</string>
+		<string>zerion</string>
+		<string>rainbow</string>
+		<string>spot</string>
```
arein, line 12: @ignaciosantise how did you solve this?

ignaciosantise: dapps will have to modify their info.plist/android manifest manually to support the feature: https://docs.walletconnect.com/2.0/advanced/walletconnectmodal/about#for-ios

PR #1002, opened 2023-07-31
[Dispatcher] Handle fallback to .org

alexander-lsvk:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectRelay/Dispatching.swift
+++ b/Sources/WalletConnectRelay/Dispatching.swift
@@ -20,6 +20,8 @@ final class Dispatcher: NSObject, Dispatching {
     private let logger: ConsoleLogging
     
     private let defaultTimeout: Int = 5
+    /// The property is used to determine whether relay.walletconnect.org will be used
+    /// in case relay.walletconnect.com doesn't respond for some reason (most likely due to being blocked in the user's location).
     private var fallback = false
     
     private let socketConnectionStatusPublisherSubject = CurrentValueSubject<SocketConnectionStatus, Never>(.disconnected)
@@ -73,10 +75,7 @@ final class Dispatcher: NSObject, Dispatching {
             .filter { $0 == .connected }
             .setFailureType(to: NetworkError.self)
             .timeout(.seconds(defaultTimeout), scheduler: concurrentQueue, customError: { .webSocketNotConnected })
-            .sink(receiveCompletion: { [weak self] result in
-                guard let self else {
-                    return
-                }
+            .sink(receiveCompletion: { [unowned self] result in
                 switch result {
                 case .failure(let error):
                     cancellable?.cancel()


Sources/WalletConnectRelay/Dispatching.swift
```diff
@@ -20,7 +20,8 @@ final class Dispatcher: NSObject, Dispatching {
     private let logger: ConsoleLogging
     
     private let defaultTimeout: Int = 5
-
+    private var fallback = false
```
llbartekll, line 23: could we describe the purpose of the variable here, so it will be easier to get context later?

Sources/WalletConnectRelay/Dispatching.swift
```diff
@@ -72,10 +73,14 @@ final class Dispatcher: NSObject, Dispatching {
             .filter { $0 == .connected }
             .setFailureType(to: NetworkError.self)
             .timeout(.seconds(defaultTimeout), scheduler: concurrentQueue, customError: { .webSocketNotConnected })
-            .sink(receiveCompletion: { result in
+            .sink(receiveCompletion: { [weak self] result in
```
llbartekll, line 76: couldn't be unowned?

Sources/WalletConnectRelay/Dispatching.swift
```diff
@@ -118,8 +126,19 @@ final class Dispatcher: NSObject, Dispatching {
...
+                self.socket.request.url = relayUrlFactory.create(fallback: fallback)
+            }
+            Task(priority: .high) {
+                await self.socketConnectionHandler.handleDisconnection()
             }
+        }
+    }
+    
+    private func handleFallbackIfNeeded(error: NetworkError) {
+        if error == .webSocketNotConnected && socket.request.url?.host == NetworkConstants.defaultUrl {
```
llbartekll, line 138: can you help me understand the logic? it will fallback if socket did not connected and url is `relay.walletconnect.com`? 
could the fallback also happen when socket did not connected from different reasons?

Sources/WalletConnectRelay/RelayURLFactory.swift
```diff
@@ -15,10 +15,10 @@ struct RelayUrlFactory {
         self.socketAuthenticator = socketAuthenticator
     }
 
-    func create() -> URL {
+    func create(fallback: Bool) -> URL {
         var components = URLComponents()
         components.scheme = "wss"
-        components.host = relayHost
+        components.host = fallback ? NetworkConstants.fallbackUrl : relayHost
```
llbartekll, line 21: does this mean that if a wallet wants to enable fallback, for all it's users regardless of their location the url will be `.org` for all of them?

llbartekll (approved):

PR #1004, opened 2023-08-02
Notify refactor debug session

llbartekll:
# Description
all notify tests are passing 
web3inbox is integrated

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectKMS/Serialiser/Serializer.swift
```diff
@@ -72,6 +73,8 @@ public class Serializer: Serializing {
 
     private func decode<T: Codable>(sealbox: Data, symmetricKey: Data) throws -> (T, Data) {
         let decryptedData = try codec.decode(sealbox: sealbox, symmetricKey: symmetricKey)
-        return (try JSONDecoder().decode(T.self, from: decryptedData), decryptedData)
+        let str = String(decoding: decryptedData, as: UTF8.self)
```
flypaper0, line 76: why we should do that?

flypaper0 (approved):

PR #1016, opened 2023-08-07
[Verify, Echo] Fallback to .org

alexander-lsvk:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectVerify/OriginVerifier.swift
+++ b/Sources/WalletConnectVerify/OriginVerifier.swift
@@ -6,6 +6,9 @@ public final class OriginVerifier {
     }
     
     private var verifyHost: String
+    /// The property is used to determine whether verify.walletconnect.org will be used
+    /// in case verify.walletconnect.com doesn't respond for some reason (most likely due to being blocked in the user's location).
+    private var fallback = false
     
     init(verifyHost: String) {
         self.verifyHost = verifyHost
@@ -29,6 +32,11 @@ public final class OriginVerifier {
             }
             return origin
         } catch {
+            if (error as? HTTPError) == .couldNotConnect && !fallback {
+                fallback = true
+                verifyHostFallback()
+                return try await verifyOrigin(assertionId: assertionId)
+            }
             throw error
         }
     }
--- a/Sources/WalletConnectVerify/VerifyClient.swift
+++ b/Sources/WalletConnectVerify/VerifyClient.swift
@@ -16,9 +16,6 @@ public actor VerifyClient: VerifyClientProtocol {
     let appAttestationRegistrer: AppAttestationRegistrer
     
     private let verifyHost: String
-    /// The property is used to determine whether verify.walletconnect.org will be used
-    /// in case verify.walletconnect.com doesn't respond for some reason (most likely due to being blocked in the user's location).
-    private var fallback = false
 
     init(
         verifyHost: String,
@@ -37,16 +34,7 @@ public actor VerifyClient: VerifyClientProtocol {
     }
 
     public func verifyOrigin(assertionId: String) async throws -> String {
-        do {
-            return try await originVerifier.verifyOrigin(assertionId: assertionId)
-        } catch {
-            if (error as? HTTPError) == .couldNotConnect && !fallback {
-                fallback = true
-                originVerifier.verifyHostFallback()
-                return try await verifyOrigin(assertionId: assertionId)
-            }
-            throw error
-        }
+        return try await originVerifier.verifyOrigin(assertionId: assertionId)
     }
     
     nonisolated public func createVerifyContext(origin: String?, domain: String) -> VerifyContext {


Sources/HTTPClient/HTTPNetworkClient.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 public actor HTTPNetworkClient: HTTPClient {
 
-    let host: String
+    var host: String
```
radeknovis, line 5:
```suggestion
    private(set) var host: String
```

radeknovis: curios if it does need to public at all?

Sources/HTTPClient/HTTPNetworkClient.swift
```diff
@@ -31,6 +31,10 @@ public actor HTTPNetworkClient: HTTPClient {
             }
         }
     }
+    
+    public func updateHost(host: String) async {
```
radeknovis, line 35: why async? is it because of HTTPNetworkClient being an `actor`?

Sources/WalletConnectEcho/Register/EchoRegisterService.swift
```diff
@@ -33,29 +33,43 @@ actor EchoRegisterService {
...
+            let response = try await httpClient.request(
+                EchoResponse.self,
+                at: EchoAPI.register(clientId: clientIdMutlibase, token: token, projectId: projectId, environment: environment, auth: echoAuthToken)
+            )
+            guard response.status == .success else {
+                throw Errors.registrationFailed
+            }
+            logger.debug("Successfully registered at Echo Server")
+        } catch {
+            throw error
```
radeknovis, line 47: Why do we need it this? Won't `throws` at function declaration be enough?

Sources/HTTPClient/HTTPError.swift
```diff
@@ -1,10 +1,27 @@
...
+public enum HTTPError: Error, Equatable {
     case malformedURL(HTTPService)
+    case couldNotConnect
     case dataTaskError(Error)
     case noResponse
     case badStatusCode(Int)
     case responseDataNil
     case jsonDecodeFailed(Error, Data)
+    
+    public static func ==(lhs: HTTPError, rhs: HTTPError) -> Bool {
```
radeknovis, line 12: nit: I wonder if you could use `if case let` pattern matching instead of conforming to Equatable. When I'm not sure how to write it I use this website https://fuckingifcaseletsyntax.com/ :D

radeknovis (approved):
Left mostly minor questions, otherwise looks good!

Sources/HTTPClient/HTTPError.swift
radeknovis: Not necessarily `if case let` works in same way as `switch`, and you don't need Equatable for the switch.

Sources/WalletConnectEcho/EchoClient.swift
```diff
@@ -3,17 +3,39 @@ import Foundation
 public class EchoClient: EchoClientProtocol {
     private let registerService: EchoRegisterService
 
+    /// The property is used to determine whether echo.walletconnect.org will be used
+    /// in case echo.walletconnect.com doesn't respond for some reason (most likely due to being blocked in the user's location).
+    private var fallback = false
```
llbartekll, line 8: will it also fallback in case user has no access to the internet?

Sources/WalletConnectEcho/EchoClient.swift
```diff
@@ -3,17 +3,39 @@ import Foundation
...
-        try await registerService.register(deviceToken: deviceToken)
+        do {
+            try await registerService.register(deviceToken: deviceToken)
+        } catch {
+            if (error as? HTTPError) == .couldNotConnect && !fallback {
+                fallback = true
+                await registerService.echoHostFallback()
+                try await registerService.register(deviceToken: deviceToken)
+            }
+            throw error
```
llbartekll, line 23: could we move the implementation down to registerService instead of handling it in the client?
so client stays as simple as possible.

llbartekll: I think it could be even done in a separate class with a single responsibility

Sources/WalletConnectVerify/OriginVerifier.swift
```diff
@@ -5,22 +5,36 @@ public final class OriginVerifier {
...
     func verifyOrigin(assertionId: String) async throws -> String {
-        let httpClient = HTTPNetworkClient(host: verifyHost)
-        let response = try await httpClient.request(
-            VerifyResponse.self,
-            at: VerifyAPI.resolve(assertionId: assertionId)
-        )
-        guard let origin = response.origin else {
-            throw Errors.registrationFailed
+        let sessionConfiguration = URLSessionConfiguration.default
+        sessionConfiguration.timeoutIntervalForRequest = 5.0
```
llbartekll, line 16: 5.0 seconds is intended for fallback correct?
what if it's already on .org, do we need a timeout then?

Sources/WalletConnectVerify/VerifyClient.swift
```diff
@@ -34,7 +37,16 @@ public actor VerifyClient: VerifyClientProtocol {
...
+        do {
+            return try await originVerifier.verifyOrigin(assertionId: assertionId)
+        } catch {
+            if (error as? HTTPError) == .couldNotConnect && !fallback {
+                fallback = true
+                originVerifier.verifyHostFallback()
+                return try await verifyOrigin(assertionId: assertionId)
+            }
+            throw error
+        }
```
llbartekll, line 49: same here could we move the implementation from the client?

llbartekll: >@llbartekll done


it's still in the client 😅

llbartekll (approved):

PR #1057, opened 2023-08-25
Set ----> OrderSet

xumoyan:
Wallets sometimes encounter a situation where authorizations do not receive a response when interacting with certain dApps. Upon investigation, it has been found that dApps retrieve the first element from the authorization array, and using a Set can alter the order of authorizations. The recommended approach is to use an OrderSet, which not only removes duplicates but also maintains the original order of authorizations.

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/WalletConnect-Package.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/WalletConnect-Package.xcscheme
@@ -482,6 +482,76 @@
                ReferencedContainer = "container:">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "WalletConnect_WalletConnectModal"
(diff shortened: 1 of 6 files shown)


flypaper0:
I think it's not safe to rely on elements order. More safe approach can be to iterate by set.

flypaper0:
Why dapps taking first element? It seems like vary optimistic implementation.

xumoyan:
Get the first item from the list obtained from [https://curve.fi/#/ethereum/swap. ](https://curve.fi/#/ethereum/swap.) Not sure if there are any other dapps to handle it. OrderSet both maintains the order and removes duplicate elements, which is more reasonable

llbartekll:
@xumoyan  can you give a specific example?

xumoyan:
Wallet: TP iOS Version
DApp: Curve

After connecting to Wallet Connect, Curve is likely unable to display the connection status. However, switching to a different network reveals that the connection is working properly.

The issue with Curve arises from accessing the first element of the array it retrieves. However, the first element is not necessarily the chain ID for Ethereum.

Testing on Android and web platforms has been successful. Only iOS is affected due to the use of a Set, which changes the order and causes this problem.

xumoyan:
@llbartekll Has this problem not recurred?

llbartekll:
no one has ever reported similar issue, so I am trying to understand.
It may also be just an incorrect integration.

can you share all the namespaces used in the session, exact requests and errors?

flypaper0:
I'm closing it due to inactivity. Looks like incorrect SDK integration from Curve side

PR #1069, opened 2023-08-30
[W3I] Redirects fix

flypaper0:


Sources/Web3Inbox/Web3InboxClient.swift
```diff
@@ -61,6 +61,10 @@ public final class Web3InboxClient {
     public func register(deviceToken: Data) async throws {
         try await notifyClient.register(deviceToken: deviceToken)
     }
+
+    public func reload() {
```
llbartekll, line 65: Should we have it as a public method?
For sample app make sense to reload but does it during integration?

flypaper0: Let leave it for now? I'm not 100% sure that logout issue totally fixed. This makes possible to recover state

PR #1087, opened 2023-09-11
[Notify] Identity refactor

flypaper0:


Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
```diff
@@ -86,8 +86,8 @@ public class NotifyClient {
         self.notifySubscriptionsChangedRequestSubscriber = notifySubscriptionsChangedRequestSubscriber
     }
 
-    public func register(account: Account, onSign: @escaping SigningCallback) async throws {
-        _ = try await identityClient.register(account: account, onSign: onSign)
+    public func register(account: Account, domain: String, isLimited: Bool, onSign: @escaping SigningCallback) async throws {
```
llbartekll, line 89: should we make isLimited=false by default?

llbartekll (approved):

Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
flypaper0: let's do

PR #1106, opened 2023-09-13
[Notify] App removed from JWT's

flypaper0:


flypaper0:
On hold until Monday

llbartekll (approved):

llbartekll (approved):
can you verify before merging that after pulling from notify-source-of-truth the tests are passing?

PR #1107, opened 2023-09-14
[Notify] Update subscriber cleanup

flypaper0:


llbartekll:
could be also remove updateSubscriptionPublisher here

PR #1109, opened 2023-09-14
[Notify] Native sample

flypaper0:


Example/WalletApp/BusinessLayer/ListingsSertice/Listings.swift
```diff
@@ -0,0 +1,45 @@
...
+            let secondary: String?
+        }
+        let shortName: String
+        let colors: Colors
+    }
+    let id: String
+    let name: String
+    let description: String
+    let homepage: String
+    let image_url: ImageURL
```
llbartekll, line 40: could we add codingkeys and use camelCase instead?

llbartekll (approved):

Example/WalletApp/BusinessLayer/ListingsSertice/Listings.swift
flypaper0: Is it worth it you think?

PR #1123, opened 2023-09-19
[Notify] Sample Subscription and Preferences screens

flypaper0:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


llbartekll (approved):
great work!
we could probably rename some classes like PushMessagesInteractor to notify

llbartekll:
also from the screen recording I can see that images are not cashed and they appear black first on every scroll up and down

PR #1124, opened 2023-09-20
[Relay] Authenticator Fix

alexander-lsvk:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectRelay/ClientAuth/ClientIdAuthenticator.swift
+++ b/Sources/WalletConnectRelay/ClientAuth/ClientIdAuthenticator.swift
@@ -4,9 +4,9 @@ public protocol ClientIdAuthenticating {
     func createAuthToken(url: String?) throws -> String
 }
 
-public struct ClientIdAuthenticator: ClientIdAuthenticating {
+public final class ClientIdAuthenticator: ClientIdAuthenticating {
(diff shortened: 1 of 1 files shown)


Sources/WalletConnectRelay/ClientAuth/ClientIdAuthenticator.swift
```diff
@@ -1,21 +1,23 @@
...
+    private var url: String
 
     public init(clientIdStorage: ClientIdStoring, url: String) {
         self.clientIdStorage = clientIdStorage
         self.url = url
     }
 
-    public func createAuthToken() throws -> String {
+    public func createAuthToken(url: String? = nil) throws -> String {
+        url.flatMap { self.url = $0 }
```
flypaper0, line 17: why we're rewriting class instance?

flypaper0: if it enough to just use and not change class state?? 
`let audience = fallbackUrl ?? url`

flypaper0: ok, let's merge

flypaper0 (approved):

PR #1153, opened 2023-10-04
Store uri into Session

arturdev:
# Description

Added `wcUri` into the Session for later use. 
Was supported in WalletConnect V1. 
Makes is easy to migrate from WalletConnect V1 from V2 for dApps, especially when dApps have to support both.

### DIFF
--- a/Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
+++ b/Sources/WalletConnectSign/Engine/Common/ApproveEngine.swift
@@ -156,6 +156,7 @@ final class ApproveEngine {
         let session = WCSession(
             topic: topic,
             pairingTopic: pairingTopic,
+            symKey: agreementKeys.sharedKey.hexRepresentation,
             timestamp: Date(),
             selfParticipant: selfParticipant,
             peerParticipant: proposal.proposer,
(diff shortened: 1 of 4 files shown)


flypaper0:
DApps should not support both because V1 was deprecated. SymKey is a sensitive information witch is stored into keychain. As we're storing sessions in UserDefaults I think it's not good idea to add this into Session object. @llbartekll what you think?

arturdev:
@flypaper0 in that case, when the dapp wants the wallet to sign a transaction, how to construct the URL to open it? 
In v1, dapps were using the WCURL for opening the wallet app (f.e. metamask://wc?{wcUrl}). 
What dapps should do in case of V2?

arturdev:
@flypaper0 but we don't connect everytime, right? The new V2 stores the connections, and on the next app run it auto connects the old sessions. So f.e. if previously you've connected Trust wallet, and killed and reopened the app, it will still be connected... So you won't call the `connect` again. But, you need to send some sign transaction f.e. But if you don't call the `connect`, you don't have the wcUri. So which URL to use to open the trust app in this case?

llbartekll:
@arturdev if I understand you correctly.
you can getSessions() from the client
the Session object has `public let peer: AppMetadata` property and it should contain enough to open the wallet from your dapp, if this is want you want to achieve

arturdev:
@llbartekll exactly. So currently `public let peer: AppMetadata` doesn't contain any wcURL which I should append to the wallet app's urlScheme in order to open it. So the only question is, how to open the wallet app (which URL to use) AFTER getting the session from `getSessions()` and not after initial connect.

arturdev:
@llbartekll Could you answer the question, please?

llbartekll:
so the thing is that wallet didn't set the redirect uri in the metadata, we will make this field required in next release.
and you will be rely on that when they upgrade.

PR #1187, opened 2023-10-17
[Sign] Dapp extend session exp

alexander-lsvk:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectSign/Sign/SignClient.swift
+++ b/Sources/WalletConnectSign/Sign/SignClient.swift
@@ -110,6 +110,8 @@ public final class SignClient: SignClientProtocol {
     private let nonControllerSessionStateMachine: NonControllerSessionStateMachine
     private let controllerSessionStateMachine: ControllerSessionStateMachine
     private let sessionExtendRequester: SessionExtendRequester
+    private let sessionExtendRequestSubscriber: SessionExtendRequestSubscriber
+    private let sessionExtendResponseSubscriber: SessionExtendResponseSubscriber
     private let appProposeService: AppProposeService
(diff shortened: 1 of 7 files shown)


Sources/WalletConnectSign/Sign/SignClient.swift
```diff
@@ -259,13 +259,18 @@ public final class SignClient: SignClientProtocol {
...
     public func extend(topic: String) async throws {
         let ttl: Int64 = Session.defaultTimeToLive
-        if sessionEngine.hasSession(for: topic) {
+        guard let session = sessionEngine.getWCSessions().first(where: { $0.topic == topic }) else {
+            throw WalletConnectError.noSessionMatchingTopic(topic)
+        }
+        if session.selfIsController {
             try await controllerSessionStateMachine.extend(topic: topic, by: ttl)
+        } else {
+            try await nonControllerSessionStateMachine.extend(topic: topic, by: ttl)
```
llbartekll, line 273: wouldn't be easier to just extract generic "sessionExtendRequester" and have an extend logic in one place?

are the implementations for a dapp and a wallet any different?

llbartekll (commented):
I would just extract the extend service as it seems like the logic is the same.

llbartekll (commented):
Could we also maybe now extract the on extend part?
now the logic is shared and we still keep request handling and responding in controller and non-controller services.
in let's say SessionExtendRequestSubscriber - that will handle the request and respond
and if needed SessionExtendResponseSubscriber that will handle the response.

llbartekll (approved):

PR #1200, opened 2023-10-27
[Notify] AppPubKey for subscription /  subscriptionChange subscriber fixed

flypaper0:


Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
```diff
@@ -132,6 +132,11 @@ public class NotifyClient {
 
 #if targetEnvironment(simulator)
 extension NotifyClient {
+
+    public var subscriptionChangedPublisher: AnyPublisher<[NotifySubscription], Never> {
```
llbartekll, line 136: didn't we want to make this one internal?

llbartekll (approved):

Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
flypaper0: I decided to make it public because we have `#if targetEnvironment(simulator)` check

PR #1202, opened 2023-10-30
Native Redirect

alexander-lsvk:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectRouter/Router/Router.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+public struct WalletConnectRouter {
+    public static func goBack(uri: String) {
+        DispatchQueue.main.async {
```
flypaper0, line 5: do we nee separate package to call `UIApplication.shared.open`?
May be just remove Router package?

Sources/WalletConnectRouter/Router/Router.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+public struct WalletConnectRouter {
+    public static func goBack(uri: String) {
```
llbartekll, line 4: what happens if the uri is an empty string for example?
should the function throw?

Example/WalletApp/PresentationLayer/Wallet/SessionRequest/SessionRequestInteractor.swift
```diff
@@ -10,6 +12,12 @@ final class SessionRequestInteractor {
                 requestId: sessionRequest.id,
                 response: .response(result)
             )
+            
+            /* Redirect */
+            let session = getSession(topic: sessionRequest.topic)
+            if let uri = session?.peer.redirect.native {
+                WalletConnectRouter.goBack(uri: uri)
```
llbartekll, line 19: how do you distinguish new router from the legacy one here?

llbartekll (approved):

PR #1203, opened 2023-10-30
[Core] Keychain migration

flypaper0:
# Description
tryMigrateAttrAccessible: 
- Check if Migration needed
- Fetching old value
- Deleting old value
- Replacing with new value

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectKMS/Keychain/KeychainStorage.swift
+++ b/Sources/WalletConnectKMS/Keychain/KeychainStorage.swift
@@ -55,7 +55,8 @@ public final class KeychainStorage: KeychainStorageProtocol {
         case errSecSuccess:
             return item as? Data
         case errSecItemNotFound:
-            return nil
+            // TODO: Replace with nil once migration period ends
+            return try tryMigrateAttrAccessible(key: key)
         default:
             throw KeychainError(status)
         }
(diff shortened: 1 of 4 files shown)


Sources/WalletConnectKMS/Keychain/KeychainStorage.swift
```diff
@@ -100,11 +101,46 @@ public final class KeychainStorage: KeychainStorageProtocol {
...
+        var item: CFTypeRef?
+        let status = secItem.copyMatching(query as CFDictionary, &item)
+
+        switch status {
+        case errSecSuccess: // Migration needed
+            guard let data = item as? Data else { return nil }
+            
+            // Fetching old value
+            var deleteQuery = buildBaseServiceQuery(for: key)
+            deleteQuery[kSecAttrAccessible] = kSecAttrAccessibleWhenUnlockedThisDeviceOnly
```
llbartekll, line 126: if read failed for `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` why delete won't

Sources/WalletConnectKMS/Keychain/KeychainStorage.swift
```diff
@@ -55,7 +55,8 @@ public final class KeychainStorage: KeychainStorageProtocol {
         case errSecSuccess:
             return item as? Data
         case errSecItemNotFound:
```
llbartekll, line 57: is it the error case we are getting when app is in background?

llbartekll (approved):
tested and it worked as expected

llbartekll (approved):

PR #1209, opened 2023-11-02
Make metadata.redirect optional

alexander-lsvk:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/WalletConnectPairing/Types/AppMetadata.swift
```diff
@@ -61,7 +61,7 @@ public struct AppMetadata: Codable, Equatable {
         description: String,
         url: String,
         icons: [String],
-        redirect: Redirect
+        redirect: Redirect?
     ) {
```
llbartekll, line 65: can we make it required in init and optional when decoding from the network

PR #1232, opened 2023-11-14
[WalletApp] Handle removed session request

alexander-lsvk:
Fix for https://walletconnect.slack.com/archives/C03S6NN8NHF/p1699699835241069

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
+++ b/Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
@@ -3,7 +3,6 @@ import Combine
 
 final class SessionEngine {
     enum Errors: Error {
-        case sessionNotFound(topic: String)
         case sessionRequestExpired
     }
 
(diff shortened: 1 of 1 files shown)


Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
```diff
@@ -281,3 +281,12 @@ private extension SessionEngine {
         onEventReceived?(topic, event.publicRepresentation(), payload.request.chainId)
     }
 }
+
+extension SessionEngine.Errors: LocalizedError {
+    var errorDescription: String? {
```
flypaper0, line 286: is it better to use WalletConnectError? we have `noSessionMatchingTopic` for example

Sources/WalletConnectSign/Engine/Common/SessionEngine.swift
```diff
@@ -281,3 +280,11 @@ private extension SessionEngine {
...
+        switch self {
+        case .sessionRequestExpired:    return "Session request expired"
```
flypaper0, line 287: It also can be moved in WalletConnectError imo

flypaper0 (approved):

PR #1237, opened 2023-11-20
[Notify] Optional icon url

flypaper0:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Example/PNDecryptionService/NotificationService.swift
+++ b/Example/PNDecryptionService/NotificationService.swift
@@ -26,7 +26,7 @@ class NotificationService: UNNotificationServiceExtension {
 
                 log("message decrypted", account: account, topic: topic, message: pushMessage)
 
-                let updatedContent = try handle(content: content, pushMessage: pushMessage, topic: topic)
+                let updatedContent = handle(content: content, pushMessage: pushMessage, topic: topic)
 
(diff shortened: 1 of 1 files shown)


Example/PNDecryptionService/NotificationService.swift
```diff
@@ -62,57 +62,62 @@ class NotificationService: UNNotificationServiceExtension {
...
+
+            incomingMessagingIntent.setImage(senderAvatar, forParameterNamed: \.sender)
+
+            let interaction = INInteraction(intent: incomingMessagingIntent, response: nil)
+            interaction.direction = .incoming
+            interaction.donate(completion: nil)
+
+            return try content.updating(from: incomingMessagingIntent)
+        } 
+        catch {
```
chris13524, line 118: This feels like a hack to wrap the whole thing in a try/catch. Can't you check for `""` and then disable the logic?

flypaper0: Execution also can failed on: 
1. downloading resource (network problem)
2. accessing device storage

Before that fix all of that cases failing message decoding

flypaper0: Nice that we found that

chris13524 (commented):
LGTM

llbartekll (approved):

PR #1252, opened 2023-12-06
[Notify] Registration api update

flypaper0:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change


Sources/Chat/ChatClient.swift
```diff
@@ -92,25 +92,34 @@ public class ChatClient {
...
+            }
 
-        return publicKey
+            try await goPublic(account: account)
+
+            return publicKey
+
+        case .rejected:
+            fatalError("Not implemented")
+        }
```
llbartekll, line 122: keep refactoring the chat is waisting of time, we can think what to do about it

llbartekll (approved):

PR #1280, opened 2024-01-18
[RPCHistory] Delete on resolve

flypaper0:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectUtils/RPCHistory/RPCHistory.swift
+++ b/Sources/WalletConnectUtils/RPCHistory/RPCHistory.swift
@@ -35,14 +35,14 @@ public final class RPCHistory {
         try? storage.get(key: recordId.string)
     }
 
-    public func set(_ request: RPCRequest, forTopic topic: String, emmitedBy origin: Record.Origin) throws {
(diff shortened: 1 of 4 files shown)


Sources/WalletConnectUtils/RPCHistory/RPCHistory.swift
```diff
@@ -95,3 +105,23 @@ public final class RPCHistory {
...
+        let thirtyDays: TimeInterval = 30*86400
+
+        for var record in records {
+            if let timestamp = record.timestamp {
+                if timestamp.distance(to: Date()) > thirtyDays {
+                    storage.delete(forKey: record.id.string)
+                }
+            } else {
+                record.timestamp = Date()
+                storage.set(record, forKey: "\(record.id)")
```
llbartekll, line 123: is it thread safe?

llbartekll (approved):

Sources/WalletConnectUtils/RPCHistory/RPCHistory.swift
flypaper0: yes, user defaults is thread safe

PR #1285, opened 2024-01-23
[Dapp] Bump W3M and include sample wallet in customWallets

radeknovis:


Example/DApp/Modules/Sign/SessionAccount/SessionAccountView.swift
```diff
@@ -210,13 +210,14 @@ struct SessionAccountView: View {
                     
                     Spacer()
                     
-                    let record = Sign.instance.getSessionRequestRecord(id: response.id)!
```
radeknovis, line 213: @flypaper0  I was getting crashes here `getSessionRequestRecord` resulting in nil. Maybe worth investigating why there was no history record when showing the response for the `personal_sign` request.

llbartekll (approved):

PR #1309, opened 2024-02-19
unregister push on logout

llbartekll:
# Description

Resolves # (issue)

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
+++ b/Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
@@ -103,10 +103,12 @@ public class NotifyClient {
     }
 
     public func unregister(account: Account) async throws {
-        try await identityClient.unregister(account: account)
+        if identityClient.isIdentityRegistered(account: account) {
+            try await identityClient.unregister(account: account)
+        }
         notifyWatcherAgreementKeysProvider.removeAgreement(account: account)
         try notifyStorage.clearDatabase(account: account)
-        Task { try await pushClient.unregister()}
+        try await pushClient.unregister()
         notifyAccountProvider.logout()
         subscriptionWatcher.stop()
     }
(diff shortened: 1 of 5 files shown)


quetool (approved):

Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
```diff
@@ -106,6 +106,7 @@ public class NotifyClient {
         try await identityClient.unregister(account: account)
         notifyWatcherAgreementKeysProvider.removeAgreement(account: account)
         try notifyStorage.clearDatabase(account: account)
+        Task { try await pushClient.unregister()}
```
chris13524, line 109: What if the unregister fails (e.g. no internet?) Will this flow be tried again or will the push registration still be present?

llbartekll: good point, I will make it throw

llbartekll: @chris13524 what code push api returns for already unregistered client on /unregister ?

chris13524: It gives the same response code regardless if the client was already unregistered or not: 200

Sources/WalletConnectPush/UnregisterService.swift
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            var request = URLRequest(url: url)
+            request.httpMethod = "DELETE"
+            request.addValue("\(pushAuthToken)", forHTTPHeaderField: "Authorization")
+
+            do {
+                let (_, response) = try await URLSession.shared.data(for: request)
+
+                guard let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode == 200 else {
```
chris13524, line 44: > 200

Instead of checking for a specific status code, can you check if it is any 2xx status code? We may want to e.g. change to 204 (no content) and all 2xx codes mean success

Sources/WalletConnectNotify/Client/Wallet/NotifyClient.swift
chris13524: > I will make it throw

I still think there is an issue. If it throws, what does caller do with that? If simply display an error message, the question to me is when will it do the unregister again?

Looks to me that you are clearing the notify storage database (and other things) before you unregister the push client. So this would mean if the way the unregister takes place is based on the notify storage, then the unregister won't happen again. I'm assuming that logout should be an atomic operation to prevent the receiving of push notifications it should not.

PR #1324, opened 2024-03-18
Make `asyncForEach` internal

mluisbrown:
# Description

Makes the `asyncForEach` function in the `Sequence` extension internal instead of `public`. The function is only used internally in two private functions, so it doesn't need to be `public`. If it's public it can clash with extensions of the same name in any projects that use this library. 

It also isn't really the purpose of this library to be vending general purpose `Sequence` extensions.

## How Has This Been Tested?

No need to test, only used in 2 private functions internal to WC SDK.

## Due Dilligence

* [ ] Breaking change

### DIFF
--- a/Sources/WalletConnectUtils/Extensions/Sequence.swift
+++ b/Sources/WalletConnectUtils/Extensions/Sequence.swift
@@ -1,7 +1,7 @@
 import Combine
 import Foundation
 
-public extension Sequence {
+extension Sequence {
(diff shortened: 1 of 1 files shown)


mluisbrown:
@llbartekll why did you close this? AFAICT this issue still exists and the fix is very simple.

llbartekll:
but it breaks the project as this method is imported by other modules

mluisbrown:
Ok, that's fair enough. You could have mentioned that in the comment for why the PR is being closed 👍

